PR #17, opened 2020-10-23
[Nep 848] Add vcType-specific metadata parsing in MetricsService

everdark:
https://replika.atlassian.net/browse/NEP-848

### Tests

+ Unit test for `MetricsSerivce` is added for Health Passport VC.
+ Separate integration test against a local metrics backend has been done for browser sdk to make sure the HP VC is parsed and saved

### TODO

- [x] Bump package version for core/browser/expo/rn
- [X] Update package lock file once `metrics-lib@0.0.15` is ready in npm registry. (https://github.com/affinityproject/affinity-metrics-lib/pull/13)


sdk/core/src/services/MetricsService.ts
```diff
@@ -1,16 +1,64 @@
...
+    const addOnMetadata = this.parseSpecific(credential)
+    return { ...baseMetadata, ...addOnMetadata }
+  }
+}
+
+// TODO: move to a single file for all vcType-specific parsers?
+class HealthPassportParser extends VcMetadataParser {
+  parseSpecific(credential: any): SpecificVcMetadada {
+    const targetResources = ['Specimen', 'Observation', 'Organization']
+    const entriesIn = credential.credentialSubject.data.fhirBundle.entry
```
rbrazhnyk, line 33: let's wrap this into try/catch, as `credential` has `type: any`

sdk/core/src/services/MetricsService.ts
```diff
@@ -1,16 +1,64 @@
...
+  }
+
+  parse(credential: SignedCredential): VcMetadata {
+    const baseMetadata = this.parseCommon(credential)
+    const addOnMetadata = this.parseSpecific(credential)
+    return { ...baseMetadata, ...addOnMetadata }
+  }
+}
+
+// TODO: move to a single file for all vcType-specific parsers?
```
rbrazhnyk, line 29: I agree, this might be a good idea to extract filters to separate files.

rbrazhnyk (commented):
Looks good to me. @DenisPopov15 @Volland WDYT?

sdk/core/src/services/MetricsService.ts
everdark: Moved to a subdir `parsers` under the `services` dir as a separate module for metadata parser.

rbrazhnyk (dismissed):

rbrazhnyk (approved):

PR #26, opened 2020-10-30
Add employee schema fields

ipatka:
We need a new type for a job offer, and in addition we need to extend the base employee role to have salary & experience letter


kunalhiray7 (commented):
We might need an attribute to store the link to the experience/relieving letter. I believe it is not there in the base Organization/Employee role.

sdk/vc-data/src/data/employment/v1.ts
```diff
@@ -14,12 +15,24 @@ import {
...
+
 type PersonEmployeeRoleEV1Mixin = CreateThing<
   'PersonEmployeeRoleE',
   {
-    reference?: CreateExpandedThing<'ContactPoint'>
+    reference?: MaybeArray<CreateExpandedThing<'ContactPoint'>>
     skills?: MaybeArray<CreateExpandedThing<'DefinedTerm'> | string>
     worksFor: ExpandThing<OrganizationEV1>
+    offerLetter?: CreateThing<'URL'>
+    experienceLetter?: CreateThing<'URL'>
```
ipatka, line 34: @kunalhiray7

kunalhiray7 (approved):
👍

sdk/vc-data/src/data/employment/v1.ts
```diff
@@ -14,12 +15,24 @@ import {
...
+>
+
 type PersonEmployeeRoleEV1Mixin = CreateThing<
   'PersonEmployeeRoleE',
   {
-    reference?: CreateExpandedThing<'ContactPoint'>
+    reference?: MaybeArray<CreateExpandedThing<'ContactPoint'>>
     skills?: MaybeArray<CreateExpandedThing<'DefinedTerm'> | string>
     worksFor: ExpandThing<OrganizationEV1>
+    offerLetter?: CreateThing<'URL'>
```
JaceHensley, line 33: This should be `CreateExpandedThing<'URL'> | string`, otherwise you'd only be able to do:

```
offerLetter: {'@type': 'URL'}
```

without any other fields or just using the URL directly.

(same for `experianceLetter`)

JaceHensley (changes requested):
just two small changes :) looks great!

sdk/vc-data/src/data/employment/v1.ts
kunalhiray7: 👍 
made the requested changes.

kunalhiray7 (approved):
👍

JaceHensley (approved):

kunalhiray7 (approved):

PR #38, opened 2020-11-11
[NEP-1074] implement VC_VERIFIED/VC_SIGNED metrics at common affinity module

everdark:
To update to use common-lib for metrics sending (for `VC_VERIFIED` and `VC_SIGNED`), corresponding changes will be required on those who using methods from class `Affinity`/`Affinidi` directly. It includes:

- [ ] Cloud wallet: https://github.com/affinityproject/cloud-wallet-api/pull/32
- [ ] HP Verifier gateway: https://github.com/affinityproject/health-passport-verifier-gateway/pull/24
- [ ] Caregiver gateway: https://github.com/affinityproject/caregiver-gateway/pull/25

This is not considered as breaking changes but without updating the constructor by default events will only send to staging metrics backend.

Unit test has been run against local metrics backend to ensure events are properly saved.


common-libs/common/src/Affinity.ts
```diff
@@ -238,6 +273,13 @@ export class Affinity {
         return { result: false, error }
       }
 
+      // send VC_VERIFIED* metrics when verification is successful
+      // TODO: also record failed verification?
+      this._sendVCVerifiedMetric(credential, parse(result.data.holder.id).did)
+
+      // TODO: how to get the verifier did?
+      this._sendVCVerifiedPerPartyMetric(credential, verifierDid)
```
DenisPopov15, line 281: @everdark there is no way how to pull verifier did here - this event could not be tracked on this level and still left only on SDK side

common-libs/common/src/Affinity.ts
```diff
@@ -238,6 +273,13 @@ export class Affinity {
         return { result: false, error }
       }
 
+      // send VC_VERIFIED* metrics when verification is successful
+      // TODO: also record failed verification?
```
DenisPopov15, line 277: @everdark yeah, also though about this yesterday - Lets right now do as we did - just on success, but could you please create a ticket to update things (probably its will require updates on the event structure, etc) to send this event on both cases, and save status (verified/not verified) as well. This task will be low priority, but something for us to not forget when will have time

common-libs/common/src/Affinity.ts
```diff
@@ -18,12 +19,46 @@ if (!fetch) {
...
+  private readonly _metricsUrl: string
+  private readonly _metricsService: any
   private readonly _digestService: any
+  protected _component: EventComponent
 
   constructor(options: AffinityOptions = {}) {
     this._apiKey = options.apiKey
     this._registryUrl = options.registryUrl || DEFAULT_REGISTRY_URL
+    this._metricsUrl = options.metricsUrl || DEFAULT_METRICS_URL
+    this._component = options.component
```
DenisPopov15, line 31: @everdark what will be if component will be not passed? I believe we should use some default one for such cases

common-libs/common/src/Affinity.ts
everdark: Actually I already created a ticket for this: 
https://replika.atlassian.net/browse/NEP-851

common-libs/common/src/Affinity.ts
everdark: Agreed.

common-libs/common/src/Affinity.ts
everdark: 👍

common-libs/common/src/Affinity.ts
```diff
@@ -274,6 +322,17 @@ export class Affinity {
         controller: await didDocumentService.buildDidDocument(),
       }),
     })
+
+    // send VC_SIGNED event
+    const isTestEnvironment = process.env.NODE_ENV === 'test'
```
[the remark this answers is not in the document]
everdark: i only found this under common-libs dir:

```
❯ git grep -w process.env
common-libs/common/src/Affinity.ts:      const isTestEnvironment = process.env.NODE_ENV === 'test'
common-libs/common/src/Affinity.ts:    const isTestEnvironment = process.env.NODE_ENV === 'test'
common-libs/common/src/shared/profiler/DefaultProfilerActivator.ts:  if (process && process.env) {
common-libs/common/src/shared/profiler/DefaultProfilerActivator.ts:    return Object.keys(process.env).findIndex((i) => profilerVars.includes(i)) !== -1
```

First two lines are mine. Not sure the effect of the profiler though.
So if we want to avoid sending metrics when it is running test, what will be the go-to approach here at common-lib?

DenisPopov15 (approved):

PR #40, opened 2020-11-12
chore: Pass only env and api key in options for tests. Bump to 3.2.2

rbrazhnyk:


DenisPopov15 (dismissed):

DenisPopov15 (approved):

DenisPopov15 (approved):

rbrazhnyk:
@DenisPopov15, you asked to remove duplicates and keep helpers in tests in CORE only. I've bumped into multiple build issues (now tests folders are not compling at all, but if we do, to share those helper, there are a bunch of errors to deal with).

I propose to make that change as a separate story.

DenisPopov15 (dismissed):

PR #53, opened 2020-11-30
add court record search schema

ipatka:


sdk/vc-data/src/data/courtRecordSearch/v1.ts
```diff
@@ -0,0 +1,92 @@
...
+} from '../util'
+
+// Helper Types
+
+export type CourtRecordSearchQueryV1 = CreateThing<
+  'CourtRecordSearchQuery',
+  {
+    parent: PersonEV1
+    spouse?: PersonEV1
+    address: CreateExpandedThing<'PostalAddress'>
```
JaceHensley, line 21: I do still like the idea of a `PostalAddressRole` to contain the meta data about an address:

```ts
type PostalAddress = CreateExpandedThing<'PostalAddress'>
type PostalAddressRoleMixin = CreateThing<'PostalAddressRole', {
  address: PostalAddress
}>
type PostalAddressRole = ExtendThing<PostalAddressRoleMixin, CreateExpandedThing<'Role'>>

export type CourtRecordSearchQueryV1 = CreateThing<
  'CourtRecordSearchQuery',
  {
    parent: PersonEV1
    spouse?: PersonEV1
    address: MaybeArray<PostalAddressRole | PostalAddress>
  }
>
```

sdk/vc-data/src/data/courtRecordSearch/v1.ts
```diff
@@ -0,0 +1,92 @@
...
+  createVCContextEntry,
+  CreateExpandedThing,
+} from '../util'
+
+// Helper Types
+
+export type CourtRecordSearchQueryV1 = CreateThing<
+  'CourtRecordSearchQuery',
+  {
+    parent: PersonEV1
```
JaceHensley, line 19: Enforcing a `parent` seems a little too strict, but I'm also not sure what a "court record" is exactly so maybe it only applied to minors?

sdk/vc-data/src/data/courtRecordSearch/v1.ts
ipatka: yup we discussed yesterday but I'll add comment here too. addressStatus is being used to describe the query rather than the address

sdk/vc-data/src/data/courtRecordSearch/v1.ts
ipatka: this is to look up if a person has appeared in court or has any criminal history. I'll make parent optional here. it's just required by Idfy for their search

JaceHensley (dismissed):

kunalhiray7 (approved):
👍

DenisPopov15 (dismissed):

PR #58, opened 2020-12-07
add education schema

ipatka:


sdk/vc-data/src/data/education/v1.test.ts
```diff
@@ -0,0 +1,100 @@
...
+      },
+      context: getVCEducationPersonV1Context(),
+    })
+
+    expect(expanded).toMatchInlineSnapshot(`
+      Object {
+        "@id": "urn:uuid:9b1deb4d-3b7d-4bad-9bdd-2b0d7b3dcb6d",
+        "@type": Array [
+          "https://www.w3.org/2018/credentials#VerifiableCredential",
+          "https://schema.affinity-project.org/EducationCredentialPersonV1",
```
launganik, line 33: if this is easily configurable, shall we switch to affinidi?

ipatka: it might lead to some inconsistencies in VC validation to switch so I'd prefer not to at the moment. perhaps when we move the schema management elsewhere and make these URLs live

launganik (approved):

PR #66, opened 2020-12-21
chore: Support Open Attestation documents

rbrazhnyk:


sdk/core/src/CommonNetworkMember.ts
```diff
@@ -1522,8 +1522,12 @@ export class CommonNetworkMember {
     const credentialTypes = []
 
     for (const credential of credentialRequirements) {
-      const type = credential.type[1] // seems its always next structure type: ['Credential', 'ProofOfEmailCredential|....']
-      credentialTypes.push(type)
+      const isW3cCredential = credential.type
```
DenisPopov15, line 1525: @rbrazhnyk lets have isW3cCredential as method or helper and use it in each proper places, so we will have one place which we will need to change when will need change definition on how to detect is that w3c credential or not

PR #78, opened 2021-03-30
add pagination to WalletStorageService when accessing vault

Yengas:
## Pagination
This PR adds pagination to **WalletStorageService** to optimize performance when accessing the bloom vault. Bumped core SDK version. And updated Browser/Expo/ReactNative SDKs to use the new version.

GetCredentials / DeleteCredential use-cases in Browser/Expo/ReactNative SDKs are updated to use the new pagination, optionally.

### Pagination Logic
Bloom Vault's start/end dates are inclusive. So `/data/0/99` means retrieving the first 100 credentials. New added skip/limit parameters need to be 0 and 100 respectively to retrieve these first 100 records. By default, these parameter defaults are 0/100. Some other examples:

```
{} = '/data/0/99'
{ limit: 5 } = '/data/0/4'
{ skip: 1 } = '/data/1/100'
{ skip: 1, limit: 1 } = '/data/1/1' 
```

### Additional
- fix did anchor metrics failing breaks flow
- fix AffinityWallet returning CommonNetworkMember for Browser/Expo/React Native

### DIFF
--- a/sdk/browser/src/AffinityWallet.ts
+++ b/sdk/browser/src/AffinityWallet.ts
@@ -178,6 +178,23 @@ export class AffinityWallet extends CoreNetwork {
   }
 
   /**
+   * @description Retrieve only the credential at given index
+   * @param credentialIndex - index for the VC in vault
+   * @returns a single VC
+   */
+  async getCredentialByIndex(credentialIndex: number): Promise<any> {
+    const paginationOptions: FetchCredentialsPaginationOptions = { skip: credentialIndex, limit: 1 }
+    const blobs = await this.walletStorageService.fetchEncryptedCredentials(paginationOptions)
+
+    if (blobs.length < 1) {
+      throw new SdkError('COR-14')
+    }
+
+    const decryptedCredentials = await this.walletStorageService.decryptCredentials(blobs)
+    return decryptedCredentials[0]
+  }
+
+  /**
    * @description Finds the given credentialShareRequestToken by searching all of your credentials
    * If a token is not given, only returns the given subset of the credentials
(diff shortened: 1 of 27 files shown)


sdk/browser/src/AffinityWallet.ts
```diff
@@ -201,31 +211,46 @@ export class AffinityWallet extends CoreNetwork {
...
-      } else {
-        throw error
-      }
+    if (credentialIndex) {
+      return this.deleteCredentialByIndex(credentialIndex)
+    }
+
+    let allBlobs: any[] = []
+
+    for await (const blobs of this.walletStorageService.fetchAllEncryptedCredentialsInBatches()) {
```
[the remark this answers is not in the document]
Yengas: Talked over this in DM. We decided to keep this as an async generator, therefore we need the `for` loop.

sdk/browser/src/AffinityWallet.ts
```diff
@@ -233,7 +250,7 @@ export class AffinityWallet extends CoreNetwork {
    * @param id - id of the credential
    * @param credentialIndex - credential to remove
    */
-  async deleteCredential(id: string, credentialIndex?: any): Promise<void> {
+  async deleteCredential(id: string, credentialIndex?: string): Promise<void> {
```
DenisPopov15, line 252: @Yengas here credentials index suppose to be number, no?

Volland, line 252: why we use string for index? Should it be a number?

DenisPopov15 (approved):

DenisPopov15 (approved):

PR #85, opened 2021-04-27
foundation 858 custom messages

Volland:


sdk/core/src/services/CustomMessageTemplatesService.ts
```diff
@@ -0,0 +1,37 @@
...
+  private readonly _serviceBaseUrl: string
+  constructor(options: any) {
+    this._options = options
+    this._serviceBaseUrl = options.keyStorageUrl || STAGING_KEY_STORAGE_URL
+  }
+  async storeTemplate(input: CreateMessageTemplateInput): Promise<void> {
+    const url = `${this._serviceBaseUrl}/api/v1/message-templates/storeTemplate`
+
+    const api = new API(null, null, null, this._options)
+    console.log(this._options)
```
rbrazhnyk, line 18: -

sdk/core/test/integration/otp/index.test.ts
```diff
@@ -311,6 +312,25 @@ describe('CommonNetworkMember (flows that require OTP)', () => {
...
+    const stmp = `${Date.now()} stmp`
+    const messageParameters: MessageParameters = {
+      message: `Your verification code is ${stmp} {{CODE}}.`,
+      subject: `${stmp} {{CODE}}.`,
+    }
+    const fullOptions = getOptionsForEnvironment(true)
+
+    const token = await CommonNetworkMember.passwordlessLogin(cognitoUsername, fullOptions, messageParameters)
+
+    // NOTE: wait for 180s before providing the answer
```
rbrazhnyk, line 326: Why do we need to wait for 3 minutes?

rbrazhnyk (approved):

sdk/core/test/integration/otp/index.test.ts
Volland: it is a copy past of Otp from other tests. I think without delay otp helper give unreliable results

sdk/core/src/services/CustomMessageTemplatesService.ts
DenisPopov15: @Volland yeap, please remove console.log here

sdk/core/test/integration/otp/index.test.ts
DenisPopov15: @Volland so `DELAY` value is 1000, this means that its will wait 1s and not 180s, so not really clear why you added such comment, please remove it. Also waiting in tests 3 minutes its toooooo long, and could not be accepted

sdk/core/test/integration/otp/index.test.ts
```diff
@@ -311,6 +312,25 @@ describe('CommonNetworkMember (flows that require OTP)', () => {
...
+    await wait(DELAY)
+    const otp = await getOtp()
+    console.log(otp)
```
DenisPopov15, line 329: @Volland please remove console.log, and add proper asserts to this test, cause its does not check nothing so far

sdk/core/test/integration/otp/index.test.ts
```diff
@@ -311,6 +312,25 @@ describe('CommonNetworkMember (flows that require OTP)', () => {
...
+    // expect(otp.startsWith(stmp)).to.equal(true)
+    await CommonNetworkMember.completeLoginChallenge(token, otp, options)
```
DenisPopov15, line 331: @Volland also I believe we want to have assert here as well, to check that result is instance of sdk, etc

sdk/core/src/CommonNetworkMember.ts
```diff
@@ -632,6 +635,16 @@ export class CommonNetworkMember {
 
     const { userPoolId, clientId } = CommonNetworkMember.setEnvironmentVarialbles(options)
 
+    if (messageParameters) {
+      const customMessagesTemplateService = new CustomMessageTemplatesService(options)
```
inga-affinidi, line 639: Is it "custom messages" or "custom message"? :)

inga-affinidi: And we already have `this._customMessageTemplateService` at this point, don't we?
(Which also does not seem to be used anywhere)

sdk/core/src/services/ApiService.ts
```diff
@@ -130,7 +130,7 @@ export default class ApiService {
     let jsonResponse
 
     if (status.toString().startsWith('2')) {
-      jsonResponse = await response.json()
+      jsonResponse = status.toString() === '204' ? {} : await response.json()
```
inga-affinidi, line 133: It does not seem to affect anything because we don't actually use its return value anywhere in this PR?

Volland: not really is you try to get json from empty response you fail with an error

sdk/core/test/integration/otp/index.test.ts
Volland: I want to use new  OTP service. just need to finish tests but for sure still need to tune tests

sdk/core/src/CommonNetworkMember.ts
DenisPopov15: Agreed @Volland - lets have one convention:
customMessage`s`TemplateService = new CustomMessageTemplatesService(options)

sdk/core/src/CommonNetworkMember.ts
```diff
@@ -153,6 +155,7 @@ export class CommonNetworkMember {
     this._api = new API(registryUrl, issuerUrl, verifierUrl, { accessApiKey: this._accessApiKey })
     this._walletStorageService = new WalletStorageService(encryptedSeed, password, this._sdkOptions)
     this._revocationService = new RevocationService(this._sdkOptions)
+    this._customMessageTemplateService = new CustomMessageTemplatesService( this._sdkOptions )
```
rbrazhnyk, line 158: it seems linter is not set up properly.

sdk/core/test/integration/otp/index.test.ts
```diff
@@ -311,6 +313,27 @@ describe('CommonNetworkMember (flows that require OTP)', () => {
...
+    const tag = `${fullOptions.env}_passwordlessLogin_integration`
+    const userName = TestEmailHelper.generateEmailForTag(tag)
+    const token = await CommonNetworkMember.passwordlessLogin(userName, fullOptions, messageParameters)
+    await wait(DELAY)
+    const [{ text, subject }] = await TestEmailHelper.getEmailsForTag(tag)
+    const [_, expectedStampFromMessage, __] = text.split(delimiter)
+    const [expectedStampFromSubject, otp] = subject.split(delimiter)
+    expect(expectedStampFromMessage).to.equal(stamp)
+    expect(expectedStampFromSubject).to.equal(stamp)
+    await CommonNetworkMember.completeLoginChallenge(token, otp, fullOptions)
```
rbrazhnyk, line 334: why do we need to complete login challenge as there are no expects afterwards?

rbrazhnyk (approved):

sdk/core/test/integration/otp/index.test.ts
Volland: to validate OTP code extracted from template

rbrazhnyk (approved):

rbrazhnyk (approved):

rbrazhnyk (approved):

sdk/core/src/CommonNetworkMember.ts
```diff
@@ -632,6 +635,16 @@ export class CommonNetworkMember {
 
     const { userPoolId, clientId } = CommonNetworkMember.setEnvironmentVarialbles(options)
 
+    if (messageParameters) {
+      const customMessageTemplateService = new CustomMessageTemplatesService(options)
```
inga-affinidi, line 639: But we already have `this._customMessageTemplateService` (which is not used anywhere)

sdk/core/src/services/CustomMessageTemplatesService.ts
```diff
@@ -0,0 +1,36 @@
...
+
+    const api = new API(null, null, null, this._options)
+
+    await api.execute(null, {
+      url,
+      params: input,
+      method: 'POST',
+    })
+  }
+  async deleteTemplate(input: DeleteMessageTemplateInput): Promise<void> {
```
inga-affinidi, line 25: `deleteTemplate` does not seem to be used anywhere?

(Although I would prefer if there was a single point of responsibility for these templates, so that the same service/lambda that created them would delete them)

sdk/core/test/helpers/getOtpFromTestMailApp.ts
```diff
@@ -0,0 +1,46 @@
...
+  to: string
+  tag: string
+  timestamp: number
+}
+
+export class TestEmailHelper {
+  public static generateEmailForTag = (tag: string): string =>
+    `${EmailConfiguration.TESTMAIL_NAMESPACE}.${tag}@inbox.testmail.app`
+  public static getEmailsForTag = async (tag: string): Promise<TestEmail[]> => {
+    const url = `${EmailConfiguration.TESTMAIL_URL}${EmailConfiguration.TESTMAIL_PATH}?apikey=${EmailConfiguration.TESTMAIL_API_KEY}&namespace=${EmailConfiguration.TESTMAIL_NAMESPACE}&tag=${tag}`
```
inga-affinidi, line 30: Maybe better to use `querystring.stringify` and `new URL` there, instead of building it all by hand (and not escaping the parameters properly)?

Also, what's the point of separating `TESTMAIL_URL` and `TESTMAIL_PATH` into two different config parameters? Are they going to ever be used separately?

sdk/core/test/helpers/getOtpFromTestMailApp.ts
```diff
@@ -0,0 +1,46 @@
...
+    const data = await request.json()
+    const emails = data.emails.map(
+      (m: any): TestEmail => {
+        const { from, to, subject, html, downloadUrl, tag, timestamp, text } = m
+        return { from, to, subject, html, downloadUrl, tag, timestamp, text }
+      },
+    )
+    return emails
+  }
+  public static extractOpt = async (tag: string) => {
```
inga-affinidi, line 41: This method is not used anywhere (we extract OTP manually in tests).

Also there is a mistype, it's OTP, not OPT :)

sdk/core/test/integration/otp/index.test.ts
inga-affinidi: Maybe then add `expect(response.accessToken).to.exist` to make the intent clear?

sdk/core/test/integration/otp/index.test.ts
```diff
@@ -311,6 +313,28 @@ describe('CommonNetworkMember (flows that require OTP)', () => {
...
+    const tag = `${fullOptions.env}_passwordlessLogin_integration`
+    const userName = TestEmailHelper.generateEmailForTag(tag)
+    const token = await CommonNetworkMember.passwordlessLogin(userName, fullOptions, messageParameters)
+    await wait(DELAY)
+    const [{ text, subject }] = await TestEmailHelper.getEmailsForTag(tag)
+    // eslint-disable-next-line no-unused-vars
+    const [_, expectedStampFromMessage, __] = text.split(delimiter)
+    const [expectedStampFromSubject, otp] = subject.split(delimiter)
+    expect(expectedStampFromMessage).to.equal(stamp)
+    expect(expectedStampFromSubject).to.equal(stamp)
```
inga-affinidi, line 334: Maybe it would be more clear to extract the OTP code from subject, and then verify that
```
expect(text).to.equal(messageParameters.text.replace('{{CODE}}', otpCode));
expect(subject).to.equal(messageParameters.subject.replace('{{CODE}}, otpCode));
```
?

But it's up to you; that's just a suggestion.

sdk/core/test/helpers/getOtpFromTestMailApp.ts
```diff
@@ -0,0 +1,46 @@
...
+    const [{ text, html }] = await TestEmailHelper.getEmailsForTag(tag)
+    const matches = (html || text || '').match(/\d{6}/)
```
inga-affinidi, line 43: Not sure that's a good idea to extract codes like that. And it is not compatible with tests in this PR:

```
>     const stamp = `${Date.now()} stamp`
undefined
>     const delimiter = ':'
undefined
>     const messageParameters = {
...       message: `Your verification code is ${delimiter}${stamp}${delimiter}{{CODE}}.`,
...       subject: `${stamp}${delimiter}{{CODE}}`,
...     }
undefined
> const text = messageParameters.message.replace('{{CODE}}', '123456')
undefined
> const matches = text.match(/\d{6}/)
undefined
> matches[matches.length-1]
'161968'
```

inga-affinidi (changes requested):

sdk/core/test/helpers/getOtpFromTestMailApp.ts
Volland: ))

sdk/core/test/integration/otp/index.test.ts
Volland: On this stage we do not have {{code}} we have real otp

sdk/core/test/helpers/getOtpFromTestMailApp.ts
Volland: We need to check not just code but the somehow the custom message that we pass as a template. so that I try to bake some indicator to a string with delimiters

sdk/core/test/integration/otp/index.test.ts
inga-affinidi: I mean we can:
1) Extract OTP code;
2) Substitute it into our templates;
3) Check that the actual message and subject match the expected ones.

Instead of extracting stamp and checking that stamp matches (ignoring everything else).

sdk/core/src/services/CustomMessageTemplatesService.ts
Volland: I found a better way with lambda to clean up message

sdk/core/test/helpers/getOtpFromTestMailApp.ts
inga-affinidi: Not sure how is this comment related to `extractOtp` method?

sdk/core/src/services/CustomMessageTemplatesService.ts
inga-affinidi: So no need for this method here then?

sdk/core/test/helpers/getOtpFromTestMailApp.ts
Volland: Let's keep stuff simple. I use Oleg's implementation as a reference. It is nice to have npm module in a future

inga-affinidi: But with the proposed changes it would still be simple:

```
const urlObject = new URL(EmailConfiguration.TESTMAIL_URL)
urlObject.path = EmailConfiguration.TESTMAIL_PATH
urlObject.search = querystring.stringify({
  apiKey: EmailConfiguration.TESTMAIL_API_KEY,
  namespace: TESTMAIL_NAMESPACE,
  tag,
}
const url = urlObject.href
```

(Or without a second line, if we merge TESTMAIL_URL and TESTMAIL_PATH into a single string in config)

sdk/core/test/helpers/getOtpFromTestMailApp.ts
```diff
@@ -0,0 +1,46 @@
...
+    const data = await request.json()
+    const emails = data.emails.map(
+      (m: any): TestEmail => {
+        const { from, to, subject, html, downloadUrl, tag, timestamp, text } = m
+        return { from, to, subject, html, downloadUrl, tag, timestamp, text }
+      },
+    )
+    return emails
+  }
+  public static extractOtp = async (tag: string) => {
```
inga-affinidi, line 41: This method still does not seem to be used anywhere...

DenisPopov15 (approved):

rbrazhnyk (approved):

launganik (approved):

sridharan21994 (approved):

PR #87, opened 2021-05-04
Fix npm dependencies for correct production build

sheremet:
- Updated placement of the npm dependencies in package.json according to Dev or production build purpose


DenisPopov15 (dismissed):

rbrazhnyk (approved):

inga-affinidi:
While we're at it, could you please also check if there are any unused packages remaining?
```
npx depcheck
```

inga-affinidi:
> Updated versions of the vulnerable npm modules

Unfortunately, not all of them are updated.

For example, `@affinidi/common` continues to depend on `jsonld` v2.0.2, which depends on `xmldom` v0.1.19 (released 7 years ago), which is "Deprecated due to CVE-2021-21366 resolved in 0.5.0".
Current version of xmldom is 0.6.0; current version of jsonld is 5.2.0; and jsonld does not depend on xmldom since 4.0.0.

inga-affinidi:
Regarding my first comment: here is what I get for `@affinidi/common`:
```
❯ npx depcheck
Unused dependencies
* lodash.keyby
* node-addon-api
* tslib
Unused devDependencies
* @types/lodash.keyby
* @types/zen-observable
* @types/express
* ts-node
* typedoc
Missing dependencies
* mobileRandomBytes: .\src\shared\randomBytes.ts
```
ts-node is a false positive (it is used in tests running, but since it is only mentioned in that package.json command, depcheck does not know it is actually used). But the rest seem to be unused.

For wallet-core-sdk:
```
❯ npx depcheck
Unused dependencies
* base64url
* browserify-aes
* jsonld
* jsontokens
* multihashes
* tslib
Unused devDependencies
* @types/jsonld
* @types/zen-observable
* @types/express
* ts-node
Missing dependencies
* mobileRandomBytes: .\src\shared\randomBytes.ts
* axios: .\src\services\EmailIssuerService.ts
```
(again, ts-node is a false positive)

For wallet-browser-sdk:
```
❯ npx depcheck
Unused dependencies
* @sentry/browser
* bip32
Unused devDependencies
* @types/ethereumjs-util
* @types/express
* @types/zen-observable
* compression-webpack-plugin
* ts-node
* webpack
* webpack-cli
Missing dependencies
* randombytes: .\src\services\KeysService.ts
```

common-libs/common/package.json
```diff
@@ -39,16 +39,11 @@
   },
   "homepage": "https://github.com/affinityproject/affinidi-core-sdk#readme",
   "dependencies": {
+    "@affinidi/wallet-core-sdk": "4.2.4",
```
inga-affinidi, line 42: Why does @affinidi/common now reference @affinidi/wallet-core-sdk as non-dev-dependency???

inga-affinidi:
CodeClimate Workflow / CodeClimate Trigger fails on the step of comparing package-lock.json generated with lerna to package-lock.json committed into repository.

However:

> ```
> Run actions/setup-node@v1
> /opt/hostedtoolcache/node/12.22.1/x64/bin/node --version
> v12.22.1
> /opt/hostedtoolcache/node/12.22.1/x64/bin/npm --version
> 6.14.12
> ```

It attempts to run lerna in environment with npm v6, and it generates package-lock v1 which is naturally different from package-lock v2 committed to repository (one of the objectives of this PR is migrating to npm v7).

To fix this, we need to modify actions/setup-node to also install npm@latest. Unfortunately, I do not have the required permissions; I cannot edit github actions.

common-libs/common/src/Affinity.ts
```diff
@@ -241,6 +241,7 @@ export class Affinity {
         credential,
         documentLoader,
         verifyRevocationListCredential,
+        verifyMatchingIssuers: false,
```
inga-affinidi, line 244: This change is required by breaking change in `vc-revocation-list` v3.
(`vc-revocation-list` v2 had some vulnerable dependencies)

common-libs/tiny-lds-ecdsa-secp256k1-2019/src/secp256k1signature.ts
```diff
@@ -0,0 +1,113 @@
...
+      data: verifyData,
+    })
+
+    return proof
+  }
+
+  async verifySignature({ verifyData, proof }: VerifyParams): Promise<boolean> {
+    const signature = proof[PROOF_SIGNATURE_KEY]
+    if (typeof signature !== 'string') {
+      throw new Error(`Expected proof[${PROOF_SIGNATURE_KEY}] to be string, got ${typeof signature}`)
```
inga-affinidi, line 66: Introduced this check for typescript reasons (because according to type declarations, signature can be undefined).

It does not significantly change behavior of the code because `this.verifier.verify` calls to `signature.indexOf` unconditionally, and this call will throw an error if signature is undefined.

common-libs/common/src/Affinity.ts
DenisPopov15: @inga-affinidi We have a revocation service: https://gitlab.com/affinidi/foundational/revocation-api, also have for the revocation service e2e/integration test at SDK. Please double check that revocation credentials issued with previous SDK version, still verifying OK using this updated one, if this the case, please go ahead and updated this dependency

DenisPopov15 (approved):

Volland (approved):

PR #91, opened 2021-05-20
[FTL-1525] Use testmail.app for OTP tests & additional support for message properties

vitalyrudenko-affinidi:
Changelog:
- Use [testmail.app](https://testmail.app/docs/) for all OTP integration tests in `core`, `browser`, `react-native` and `expo` packages
- Refactored `TestmailInbox` to use [livequery](https://testmail.app/docs/#waiting-for-new-email) (waiting for new email)
- Add support for `messageProperties` option in `CommonNetworkMember#changeUsername()`, `CommonNetworkMember#resendSignUpConfirmationCode()` and `CognitoService#changeUsername()`
- Fixed linter and type issues with `tiny-lds-ecdsa-secp256k1-2019` (updated version to `1.0.2`)

TODO for subsequent PRs:
- [ ] Don't use `{{CODE}}` in the **subject** of OTP emails for passwordless login's template since it's discouraged by AWS ([link 1](https://forums.aws.amazon.com/message.jspa?messageID=870189), [link 2](https://github.com/aws-amplify/amplify-js/issues/936)) and actually not supported by the default auth flows at all
- [ ] Update `create-auth-challenge` lambda to not to replace `{{CODE}}` in the **subject** of OTP emails


sdk/core/test/integration/otp/index.test.ts
```diff
@@ -1,417 +1,426 @@
...
+  { suffix }: { suffix?: string } = {},
+): [string, string, MessageParameters] => {
+  const tag = generateTag(testId, suffix)
+  const email = TestmailHelper.generateEmailForTag(tag)
+  const messageParameters: MessageParameters = {
+    message: `Your verification code is: {{CODE}}`,
+    subject: `Verification code`,
+  }
+
+  return [email, tag, messageParameters]
```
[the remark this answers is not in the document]
vitalyrudenko-affinidi: It's more convenient to use when I need new credentials for changing username, for example:
```ts
const [email, tag, messageParameters] = ...
...
const [newEmail, newTag, newMessageParameters] = ...
...
```

sdk/core/test/integration/otp/index.test.ts
```diff
@@ -1,417 +1,426 @@
...
-    expect(networkMember).to.be.an.instanceof(CommonNetworkMember)
+  describe('[with existing user]', () => {
+    let email: string
+    let tag: string
+    let messageParameters: MessageParameters
+
+    // reuse the same user for all subsequent tests to reduce OTP emails count
+    before(async () => {
+      // testId is not available in "before" hook, so we generate it separately
+      ;[email, tag, messageParameters] = prepareOtpMessageParameters(generateTestId())
```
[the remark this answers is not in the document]
vitalyrudenko-affinidi: Not sure. I also thought about it. Perhaps generating it once (where possible, like now) reduces the risk of collision even more.
It's not that important, though, since emails are only "waited" since specified timestamp.
Subjectively it just looks somewhat "cleaner" when we have an id for a test run instead of generating random string each time we need new email. And it also corresponds to the documentation of testmail.

sdk/core/test/integration/otp/index.test.ts
```diff
@@ -1,417 +1,426 @@
...
 const options: SdkOptions = getOptionsForEnvironment()
+const { env } = getOptionsForEnvironment(true)
+
+const prepareOtpMessageParameters = (
+  testId: string,
+  { suffix }: { suffix?: string } = {},
+): [string, string, MessageParameters] => {
+  const tag = generateTag(testId, suffix)
+  const email = TestmailHelper.generateEmailForTag(tag)
+  const messageParameters: MessageParameters = {
```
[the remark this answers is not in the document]
vitalyrudenko-affinidi: As long as it's not modified somewhere deep in the code, yeah, I agree.

sdk/core/test/integration/otp/index.test.ts
```diff
@@ -1,417 +1,426 @@
...
 
 const options: SdkOptions = getOptionsForEnvironment()
+const { env } = getOptionsForEnvironment(true)
+
+const prepareOtpMessageParameters = (
+  testId: string,
+  { suffix }: { suffix?: string } = {},
+): [string, string, MessageParameters] => {
+  const tag = generateTag(testId, suffix)
+  const email = TestmailHelper.generateEmailForTag(tag)
```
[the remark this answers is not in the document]
vitalyrudenko-affinidi: I actually thought about it, but since it doesn't change the complexity in the test themselves, I don't think it's better (or worse) to do it that way. I'd rather have these actions separated and have small helpers on top of the file. 
I don't like parsing things unnecessary, because it can make email generation rigid and also raise "wtf?" questions when we come back to the code later.

sdk/browser/package-lock.json
```diff
@@ -83,6 +83,22 @@
 			"integrity": "sha512-Zywpj3D/su2oZOmMLgfaQc08Aqkwkm2O8BgHTMMgZ675KakFpTYtOZSxtrl3j/+XpA0g5NPuC486C51d3JrCGA==",
 			"dev": true
 		},
+		"@affinidi/issuer-email-ses-client": {
```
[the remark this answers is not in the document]
vitalyrudenko-affinidi: I think this is the result of the `lerna clean && lerna bootstrap`

sdk/core/test/helpers/TestmailHelper.ts
```diff
@@ -0,0 +1,34 @@
...
+interface TestmailEmail {
+  subject: string
+  html?: string
+  text?: string
+}
+
+// NOTE: Don't use this helper for mere email generation,
+//       since there are limits for emails sent per month
+//       Use generateEmail() for random email generation
+export class TestmailHelper {
```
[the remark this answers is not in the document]
vitalyrudenko-affinidi: How would you name the file?

sdk/core/test/integration/otp/index.test.ts
```diff
@@ -1,417 +1,426 @@
...
+  const messageParameters: MessageParameters = {
+    message: `Your verification code is: {{CODE}}`,
+    subject: `Verification code`,
+  }
+
+  return [email, tag, messageParameters]
+}
+
+const waitForOtpCode = async (tag: string, timestampFrom?: number): Promise<string> => {
+  const { text, html } = await TestmailHelper.waitForNewEmail(tag, timestampFrom)
```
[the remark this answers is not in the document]
vitalyrudenko-affinidi: `livequery` immediately returns the result if there is at least one matching email in the box, so your proposed logic might look too complicated in the actual code
I think we should concern ourselves with such questions only if we have concrete problems to solve.
If clock is faulty, then there are more problems then just integration tests failing - which we will see immediately.

sdk/core/test/integration/otp/index.test.ts
```diff
@@ -1,417 +1,426 @@
...
+      const timestamp = Date.now()
+      const token = await CommonNetworkMember.passwordlessLogin(email, options, messageParameters)
+      const otpCode = await waitForOtpCode(tag, timestamp)
+
+      const commonNetworkMember = await CommonNetworkMember.completeLoginChallenge(token, otpCode, options)
+      expect(commonNetworkMember).to.be.instanceOf(CommonNetworkMember)
+    })
+
+    // requires test timeout >= 180000 ms
+    it.skip('#completeLoginChallenge should throw "COR-17" for expired OTP', async () => {
```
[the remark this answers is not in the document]
vitalyrudenko-affinidi: Sounds interesting, forgot about per-test configuration. Will try!

sdk/core/test/integration/otp/index.test.ts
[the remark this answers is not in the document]
inga-affinidi: Then, since it is synchronous, maybe it is better to move it out of the top-level `describe` and just do `const testId = generateTestId()` at the top of the module? That way all tests will be able to use it; you won't have to think whether it is available in "before" hook or not.

sdk/core/test/helpers/TestmailHelper.ts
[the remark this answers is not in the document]
inga-affinidi: testMailHelpers, for example?

sdk/core/test/integration/otp/index.test.ts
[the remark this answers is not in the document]
vitalyrudenko-affinidi: Sorry for the confusion. The line we're currently discussing is an exception (since we reuse the user account).
"testId" should be unique in each spec. Though, now that I've said it, it would be easier to generate new id every time we generate an email.
I will remove it.

sdk/core/test/integration/otp/index.test.ts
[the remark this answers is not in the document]
inga-affinidi: > livequery immediately returns the result if there is at least one matching email in the box

Which is why I said "call api with livequery=true **and timestamp=lastMessageTimestamp**" :)

> If clock is faulty, then there are more problems then just integration tests failing

It does not have to be faulty, it just has to be a couple of seconds behind or ahead of clock on their server. It happens all the time, that's why we have NTP to keep it in check :)
Generally, relying on the fact that your clock is in perfect sync with remote clock is quite a bad idea; it is not guaranteed to be.
And there is a way to avoid relying on it (by supplying testmail with timestamps they generate, instead of our timestamps).

sdk/core/test/integration/otp/index.test.ts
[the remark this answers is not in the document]
inga-affinidi: It should not be modified by the code under testing. If it is, that's a bug that should be fixed :)

sdk/core/test/integration/otp/index.test.ts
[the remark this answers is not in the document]
vitalyrudenko-affinidi: Yeah, I agree, we can utilize their timestamps :+1:

sdk/core/test/integration/otp/index.test.ts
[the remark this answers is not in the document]
inga-affinidi: > but since it doesn't change the complexity in the test themselves

But it does! Right now, tests have to concern themselves with implementation details of TestMailHelper; they have to know that in addition to email there is some "tag" used to access messages.

On the other hand, we could avoid parsing things by storing the tag in the `Mailbox` object, as proposed in other comment...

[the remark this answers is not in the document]
vitalyrudenko-affinidi: Oh, I meant the specs, not the helpers around it. In combination with your mailbox proposal, it makes sense to move all these helpers into one place. I will try to do that, thank you

sdk/core/test/integration/otp/index.test.ts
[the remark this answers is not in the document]
inga-affinidi: (In the context of another discussion there) I would suggest to do something along the lines of:
```
class Mailbox {
  readonly email: string
  private readonly tag: string
  private lastMessageTimestamp = 0

  Mailbox() {
    this.tag = generateTag()
    this.email = generateEmailFromTag()
  }

  fetchNewMessage() {
    const url = new URL(TESTMAIL_API_URL)
    url.searchParams = new URLSearchParams({
      tag: this.tag,
      timestamp_from: this.lastMessageTimestamp,
      livequery: 'true',
      ...
    )
    const response = await fetch(url)
    const { emails } = await response.json()
    const { timestamp, subject, html, text } = emails[0]
    this.lastMessageTimestamp = timestamp
    return { subject, html, text }
  }
}
```

That way we would not have to parse email addresses for tags.

sdk/core/test/integration/otp/index.test.ts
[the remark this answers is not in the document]
inga-affinidi: Isn't this the test spec? :)
https://github.com/affinityproject/affinidi-core-sdk/blob/012bcd9b27a0054030f4197512a63910d4521456/sdk/core/test/integration/otp/index.test.ts#L59-L68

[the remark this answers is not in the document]
vitalyrudenko-affinidi: Well I wanted to test "messageParameters" separately and then just use the "waitForOtpCode" method everywhere, but I see your point

sdk/core/src/services/CognitoService.ts
```diff
@@ -278,11 +278,8 @@ export default class CognitoService {
 
     const { clientId: ClientId } = this.cognitoOptions
 
-    const params = { ClientId, Username }
-
-    if (messageParameters) {
-      Object.assign(params, { ClientMetadata: messageParameters })
-    }
+    const ClientMetadata = messageParameters || {}
+    const params = { ClientId, Username, ClientMetadata }
```
DenisPopov15, line 282: @VitaliiShpital did you run all tests and their working as expected for all custom message use cases, correct ?
Wondering if nothing get broken when ClientMetadata present and empty (just thinking why it was written this way before)

sdk/core/test/integration/index.test.ts
```diff
@@ -1,5 +1,6 @@
 import 'mocha'
-
 import './env'
+
 require('./services')
 require('./NetworkMember.test')
+require('./otp/index.test')
```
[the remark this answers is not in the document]
DenisPopov15: Lets have one entry point for the test suits, then number of features works as expected (as `.only`, etc). So I vote leave it this way as it done here

Volland (dismissed):

sdk/browser/test/helpers/TestmailInbox.ts
```diff
@@ -0,0 +1,59 @@
...
+  private _email: string
+  private _lastEmailTimestamp: number
+
+  constructor({ prefix, suffix }: InboxOptions) {
+    const inboxId = cryptoRandomString({ length: 12 }) // arbitrary length
+    this._tag = `${prefix}.${inboxId}.${suffix}`
+    this._email = `${TESTMAIL_NAMESPACE}.${this._tag}@${TESTMAIL_INBOX_DOMAIN}`
+
+    // if inbox has already been used, ignore its old emails
+    this._lastEmailTimestamp = Date.now() - 60_000
```
inga-affinidi, line 34: One minute?

sdk/expo/test/helpers/TestmailInbox.ts
```diff
@@ -0,0 +1,59 @@
+import fetch from 'node-fetch'
```
inga-affinidi, line 1: Looks like it's time to move out this file to some shared package instead of copy&pasting it into different projects.

What do you think, @DenisPopov15 ?

sdk/browser/test/helpers/TestmailInbox.ts
```diff
@@ -0,0 +1,59 @@
...
+}
+
+interface Email {
+  subject: string
+  body: string
+}
+
+// NOTE: Don't use this helper for mere email generation,
+//       since there are limits for emails sent per month
+//       Use generateEmail() instead
```
inga-affinidi, line 22: I do not understand what this comment means.
Does "email" in "for mere email generation" refer to the address or to the message?
When should I use this helper and when should I use `generateEmail`?
Other developers in the future might also have problems with understanding it.

sdk/browser/test/integration/otp/index.test.ts
```diff
@@ -4,26 +4,17 @@ import '../env'
...
-
-const generateEmail = () => {
-  const TIMESTAMP = Date.now().toString(16).toUpperCase()
+import { getOptionsForEnvironment } from '../../helpers'
+import { TestmailInbox } from '../../../test/helpers/TestmailInbox'
+import { openAttestationDocument } from '../../factory/openAttestationDocument'
 
-  return `test.user-${TIMESTAMP}@gdwk.in`
-}
+const signedCredentials = require('../../factory/signedCredentials')
```
inga-affinidi, line 14: You can also use import there :)

sdk/browser/test/helpers/TestmailInbox.ts
inga-affinidi: And why would we need it at all if inboxId is cryptographically secure random? If you think there are meaningful chances of collision, then we should probably increase the length of `inboxId` instead.

sdk/browser/test/integration/otp/index.test.ts
```diff
@@ -39,174 +30,154 @@ const credentialShareRequestToken =
...
+const createInbox = () => new TestmailInbox({ prefix: env, suffix: 'otp.browser' })
 
-    const networkMember = await AffinityWallet.confirmSignUp(token, signUpOtp, options)
+describe('AffinityWallet [OTP]', () => {
+  it('Save Open Attestation credential and #deleteCredential scenario', async () => {
+    const inbox = createInbox()
+    const password = COGNITO_PASSWORD
 
-    let credentials
+    const signUpToken = await AffinityWallet.signUp(inbox.email, password, options, messageParameters)
```
inga-affinidi, line 52: By the way, since we're here: @DenisPopov15 , do we really want to use `TEST_SECRETS.COGNITO_PASSWORD` here instead of generating a new password every time? 🤔
(This code was there for a long time, so the comment is not related to this PR)

sdk/core/test/integration/index.test.ts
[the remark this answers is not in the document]
inga-affinidi: @DenisPopov15 , I'm not sure what you mean? What is the problem with `.only`?

When there are multiple test files, and only one test in one of them is marked as `.only`, only that one test is run. (It already works this way in cloud-wallet-api)

sdk/core/test/unit/services/CognitoService.test.ts
```diff
@@ -711,16 +712,35 @@ describe('CognitoService', () => {
   })
 
   describe('#changeUsername', () => {
-    it(successPathTestName, async () => {
-      const error = { code: USER_NOT_FOUND_EXCEPTION }
+    describe('[username is not taken]', () => {
+      beforeEach(() => {
+        stubMethod(INITIATE_AUTH, null, { code: USER_NOT_FOUND_EXCEPTION })
```
inga-affinidi, line 717: Shouldn't it be `beforeAll`?

sdk/expo/test/helpers/TestmailInbox.ts
inga-affinidi: But also I don't see if this file is referenced anywhere in `expo`?

sdk/core/src/services/CognitoService.ts
vitalyrudenko-affinidi: Yeah nothing have changed with this tweak. I've actually copy-pasted this from other method in this class.

sdk/expo/test/helpers/TestmailInbox.ts
vitalyrudenko-affinidi: I also don't like copy-pasting, but I'm not sure where to put it. I'd rather have some lib which has useful utilities, including tests -- and would put it there.
And it it referenced in `expo`, here:
https://github.com/affinityproject/affinidi-core-sdk/blob/cba04d447eb3720320f29b89ac39c478dff58d41/sdk/expo/test/integration/otp/index.test.ts#L12

sdk/browser/test/helpers/TestmailInbox.ts
vitalyrudenko-affinidi: We don't need it cryptographically secure random, I just didn't know what else to use in this case.
Simplest solution for me was to use this lib.

One minute is arbitrary. Just in case there will ever be a collision, avoid situation where `livequery` immediately resolves with a very old email.

sdk/browser/test/helpers/TestmailInbox.ts
vitalyrudenko-affinidi: We need to avoid using this helper just for email generation, unless we want to receive emails and read them.
It's because `testmail` is a paid product, and allows no more than 10,000 mails per month for our current plan.
So for random email generation (and we don't care about the mails) we can use `generateEmail()`.
If you understood me, it would be nice if you helped me to shape this comment into a useful state.

sdk/browser/test/integration/otp/index.test.ts
vitalyrudenko-affinidi: It fails and just didn't want to mess more than I already did :D
Says "it's not a module" or something like that.

sdk/core/test/unit/services/CognitoService.test.ts
vitalyrudenko-affinidi: No, because it restores in `afterEach`:
https://github.com/affinityproject/affinidi-core-sdk/blob/65ed77a4e28177ae727fe638be9932073bb67f18/sdk/core/test/unit/services/CognitoService.test.ts#L54

sdk/browser/test/helpers/TestmailInbox.ts
inga-affinidi: But there should be no collision. And if we think there might be, it's better to increase the `inboxId` length to avoid them.

sdk/expo/test/helpers/TestmailInbox.ts
inga-affinidi: We might just create some tiny package, like `@affinidi/testmail-client`

sdk/browser/test/helpers/TestmailInbox.ts
vitalyrudenko-affinidi: Are there cons of doing it the way I did here?

sdk/browser/test/helpers/TestmailInbox.ts
inga-affinidi: I think I start to understand it now.

But if no mail messages are sent, then it does not affect the remaining quota.

So we should only avoid using TestmailInbox in cases when:
* We need an email address; and
* Some messages will be sent to this address; and
* We don't care about these messages and will never read them.

Right?

I'm not sure it is such a common scenario for us to guard against it.

sdk/expo/test/helpers/TestmailInbox.ts
vitalyrudenko-affinidi: Perhaps it can be done as a separate "polishing" task? Because I'm not sure if it's that much of a problem right now (not sure if we gonna change this code any time soon) and I'd rather focus on more important tasks.
@DenisPopov15 what do you think?

sdk/browser/test/integration/otp/index.test.ts
inga-affinidi: Ah, of course, because `signedCredentials` is indeed not a modern module for some reason: https://github.com/affinityproject/affinidi-core-sdk/blob/master/sdk/browser/test/factory/signedCredentials.ts#L67

OK, let's leave it like that in this PR

sdk/browser/test/helpers/TestmailInbox.ts
vitalyrudenko-affinidi: If I'm not mistaken, email is always sent in `signIn`/`signUp` methods, that are also tested in `NetworkMember.test.ts`. I haven't gone deep to actually test that, but I just wanted to point out that I'm discouraging usage if this helper (and testmail in general) for mere email generation. It doesn't make sense either way, since this service should only be used for testing emails.

sdk/browser/test/helpers/TestmailInbox.ts
inga-affinidi: It introduces a magic constant and makes intent harder to follow, ihmo.
But if you feel it's better that way, then let's leave it that way.

sdk/core/test/unit/services/CognitoService.test.ts
inga-affinidi: Oh. That's really hard to follow, because `afterEach` is on another level.
But since it was there before, let's leave it as it was.

sdk/expo/test/helpers/TestmailInbox.ts
inga-affinidi: > And it it referenced in expo, here:

GitHub decided that I don't need to be able to find it 🙃

>  sdk/expo/test/integration/otp/index.test.ts 
> Load diff
> Large diffs are not rendered by default.

Thank you for the tip

inga-affinidi: > Because I'm not sure if it's that much of a problem right now (not sure if we gonna change this code any time soon)

AFAIK we're going to migrate our other projects to use testmail for testing, so it would be really useful to extract it into a separate package.
(And to not have three copies of the same file in affinidi-core-sdk.)

But priorities are up to @DenisPopov15

vitalyrudenko-affinidi: image

vitalyrudenko-affinidi: > > Because I'm not sure if it's that much of a problem right now (not sure if we gonna change this code any time soon)
> 
> AFAIK we're going to migrate our other projects to use testmail for testing, so it would be really useful to extract it into a separate package.
> (And to not have three copies of the same file in affinidi-core-sdk.)
> 
> But priorities are up to @DenisPopov15

Oh, then it would be useful to extract this, indeed. I need guidance, though, in were to put it and how.
Separate small package (`@affinidi/testmail-client`) that you suggested sounds good to me.

sdk/browser/test/helpers/TestmailInbox.ts
inga-affinidi: > If I'm not mistaken, email is always sent in signIn/signUp methods, that are also tested in NetworkMember.test.ts.

If a new user is signed up by email address, then amazon does indeed send the message. However, in that case we would probably also need to get the confirmation code in order to move forward.

What I'm trying to say is:
* If we need to get the message, we should use TestmailInbox;
* If we just need some email address and no messages are getting sent, we should not use TestmailInbox (just because it was created for an entirely different purpose, as it name reflects); but using it will do no harm;
* And the cases where we need an email address, and some messages are getting sent to it, but we don't really care about their content, are not that common, and even if we use TestmailInbox there, it will not do much harm (maybe it will increase total number of messages by 10% or something like that). Even then, it is probably easier to use TestmailInbox there from the standpoint of code uniformity; and anyway, where else do we get a temporary address? Sending some test messages to testmail and some other to gdwk.in does not seem to be a good idea.

vitalyrudenko-affinidi: Still, I want to separate "testmail" inboxes and some random email generation.
Even 10% is huge amount, I've tried to minimize emails sent as much as I could, so I'm sticking to that - it allows us to keep using the cheaper plan of testmail, and also I don't think we should use a whole service helper just for an email generation.

> Sending some test messages to testmail and some other to gdwk.in does not seem to be a good idea.

We can just use something like `example.com`, it's not a problem.

sdk/browser/test/integration/otp/index.test.ts
```diff
@@ -39,174 +29,154 @@ const credentialShareRequestToken =
...
-describe('AffinityWallet (flows that require OTP)', () => {
-  it('Save Open Attestation credential and #deleteCredential scenario', async () => {
-    const cognitoUsername = generateEmail()
+const messageParameters: MessageParameters = {
+  message: `Your verification code is: {{CODE}}`,
+  subject: `Verification code`,
+}
 
-    const token = await AffinityWallet.signUp(cognitoUsername, cognitoPassword, options)
+const waitForOtpCode = async (inbox: __dangerous.TestmailInbox): Promise<string> => {
```
inga-affinidi, line 39: Why is it in __dangerous part of the main code rather than in tests?

@DenisPopov15 , can we move TestmailInbox out into a separate npm package, please?

vitalyrudenko-affinidi: It's an okay-ish start. At least it will be easier to move to a separate package, since there is no copy paste. For a separate package I think there must be a separate task.

sdk/browser/test/integration/otp/index.test.ts
vitalyrudenko-affinidi: Created a ticket: https://replika.atlassian.net/browse/FTL-1563

Volland (approved):

DenisPopov15 (approved):

PR #97, opened 2021-05-27
Optional DTO fields are marked as TypeScript optional fields 

inga-affinidi:


DenisPopov15 (approved):

Volland (dismissed):

sdk/core/src/dto/shared.dto.ts
```diff
@@ -207,11 +206,11 @@ export class SignedCredential {
 
   @IsOptional()
   @IsISO8601()
-  issuanceDate: string
+  issuanceDate?: string
```
vitalyrudenko-affinidi, line 209: (a question)
Some of these fields are required by the W3C spec, but why are they optional here?
Can signed credential be incomplete?

vitalyrudenko-affinidi (approved):

DenisPopov15 (approved):

PR #102, opened 2021-06-14
Updated getCredentials to return all credentials (browser SDK)

inga-affinidi:


sdk/browser/src/AffinityWallet.ts
```diff
@@ -195,8 +195,7 @@ export class AffinityWallet extends CoreNetwork {
    * @returns array of ids for corelated records
    */
   async saveCredentials(data: any, storageRegion?: string): Promise<any> {
-    const encryptedCredentials = await this.walletStorageService.encryptCredentials(data)
-    const result = await this.saveEncryptedCredentials(encryptedCredentials, storageRegion)
+    const result = await this.walletStorageService.saveUnencryptedCredentials(data, storageRegion)
```
DenisPopov15, line 198: @inga-affinidi this namespace is confusing. (make assumption that you will store credential in unEncrypted way). We could call it encryptAndStoreCredentials (if you want to have this as one method)

DenisPopov15: the idea of why it's separate. Cause only encryption part is different depends on the platform (browser, react, expo), all other methods - its the same, so that's why we using here this separate platform related method for encryption/decryption, and then basic functionality inherited from core services.

inga-affinidi: @DenisPopov15 , agree, I was hoping I'll be able to refactor all that stuff. Hopefully it won't take long and then it will make it to the non-draft version of that PR tomorrow. If not, I'll just rename the method.

sdk/core/src/CommonNetworkMember.ts
```diff
@@ -95,30 +96,50 @@ import {
...
+}
+type ConstructorKeys<T> = {
+  [P in keyof T]: T[P] extends new (...args: unknown[]) => unknown ? P : never
+}[keyof T]
+type OmitConstructor<T> = Omit<T, ConstructorKeys<T>>
+type DerivedThis<T> = Constructor<T> & AbstractStaticMethods<T> & OmitConstructor<typeof CommonNetworkMember>
+
 @profile()
-export class CommonNetworkMember {
+export abstract class CommonNetworkMember {
```
DenisPopov15, line 115: @inga-affinidi lets not do this class as abstract, at this point, cause I believe right now its could be used at different places at customer on backends as separate instance.
Lets mention at the CHANGE_LOG (And Probably at REAMDE as well) that its going to happen as abstract in the near future, and will be not available. And that consumers should use browser SDK where there using core SDK instead, and make proper announcement at the network_release channel about that (please post that this attention message there, when this version will be released)

common-libs/common/test/factory/CommonNetworkMember.ts
```diff
@@ -0,0 +1,26 @@
+import { EventComponent } from '@affinidi/affinity-metrics-lib'
+import { CommonNetworkMember as GenericCommonNetworkMember, __dangerous } from '@affinidi/wallet-core-sdk'
```
DenisPopov15, line 2: @inga-affinidi so we decided still keep core-sdk as dev dependency at the common lib? Ok, I am fine (its could be another task, scope to get rid from it here)

sdk/core/src/CommonNetworkMember.ts
```diff
@@ -95,30 +96,50 @@ import {
...
   private _didDocumentKeyId: string
-  protected _component: EventComponent
+  protected readonly _component: EventComponent
   protected cognitoUserTokens: CognitoUserTokens
 
-  constructor(password: string, encryptedSeed: string, options: SdkOptions = {}, component?: EventComponent) {
+  constructor(
+    password: string,
+    encryptedSeed: string,
+    platformEncryptionTools: IPlatformEncryptionTools,
```
DenisPopov15, line 139: @inga-affinidi since we going to postpone moving this class to abstract - lets move `platformEncryptionTools ` to the last parameter and if its not provided use default one (default one could be one from browser)

sdk/core/src/services/WalletStorageService.ts
```diff
@@ -524,6 +530,62 @@ export default class WalletStorageService {
       limit: limit || 100,
     }
   }
+
+  async saveUnencryptedCredentials(data: unknown[], storageRegion?: string) {
```
DenisPopov15, line 534: @inga-affinidi still see this name convention. Lets have it as encryptAndSaveCredentials (or something like that, to avoid confusion <its gives assumption that credentials will be stored at unencrypted way>)

sdk/core/src/CommonNetworkMember.ts
inga-affinidi: It has to be abstract for a reason, in order for static factories to work properly.

After this PR, `CommonNetworkMember` constructor will have `platformNetworkTools` as one of its parameters; but for the sake of simplicity, all the factory methods (e.g. `CommonNetworkMember.signUp`) do not accept `platformNetworkTools` as a parameter, instead requiring `this` constructor to be
```
new (
  password: string,
  encryptedSeed: string,
  options?: SdkOptions,
  component?: EventComponent,
) => T
```

So we could have `AffinidiWallet extends CommonNetworkMember` with its own constructor with that simpler signature (which will pass its platform tools to `CommonNetworkMember` constructor); and `AffinidiWallet.signUp` will automatically return an instance of `AffinidiWallet` with all the types declared properly, even though `AffinidiWallet` does not have a static `signUp` member.

On the other hand, it means that direct calls `CommonNetworkMember.signUp` won't work (and TypeScript won't let to make them), because `CommonNetworkMember` does not have a constructor of that simpler required type.

(We're doing some java-style OOP here, but it is worth it because it allows to remove all that complexity; a lot of methods had 3-4 identical copies before that PR in order to make everything work, and still there were some problems with types).

---

Why do customers want to use `CommonNetworkMember` directly?

If they don't need features that depend on platform encryption tools, they could always do something `common` now does: https://github.com/affinityproject/affinidi-core-sdk/blob/985b097422f40d2f9ba4a4acb2a2ca9e3be17597/common-libs/common/test/factory/CommonNetworkMember.ts

If they need these features, they will have to provide platform encryption tools anyway; and if e.g. platform encryption tools from `wallet-browser-sdk` work on their platform, why won't they just use `AffinidiWallet` from `wallet-browser-sdk`?

I understand that this is a breaking change, which is why I've bumped the version all the way up to 5.0.0. But I agree that we might want to clearly describe the migration path, if you have some specific usecases in mind.

common-libs/common/test/factory/CommonNetworkMember.ts
inga-affinidi: Yes, it should be another task, I've just reorganized code a bit in `common` tests, nothing of substance has changed.

sdk/core/src/CommonNetworkMember.ts
inga-affinidi: Allowing it to be optional and defaulting to browser is too error-prone, we might easily end up accidentally using the one from browser in expo or react-native where it is not supported.
Even if we were to make it optional, I'd say that the default should be the one we have in `common` tests (https://github.com/affinityproject/affinidi-core-sdk/blob/985b097422f40d2f9ba4a4acb2a2ca9e3be17597/common-libs/common/test/factory/CommonNetworkMember.ts) - that is, the default will just not support features that require platform tools.

However, I don't think making it optional is a good idea. There is just too many places where the constructor is called (all these static factories, like `fromSeed`, `signUp` etc). And either they will call `CommonNetworkMember` constructor (and then all these factories have to accept `platformEncryptionTools` as their optional argument, and then every client app that uses react-native SDK will have to provide its platform tools in every call to every such factory, or react-native SDK will have to override all these factories just to pass proper `platformEncryptionTools` to the base class). Or they will call the constructor of an actual non-core SDK, and that means that `CommonNetworkMember` constructor will only be called from non-core SDK constructors, and there is no need in making any of its parameters optional at all.

sdk/core/src/CommonNetworkMember.ts
DenisPopov15: Yeah, to have proper instance of classe as well, to be able get rid of this hack.
Ok, to avoid extra work for this soft version bump - then lets hihligh this at the channel and at the Chagelog explicitly with proper way what should be updated if somebody using core-sdk as it is right now

sdk/core/src/CommonNetworkMember.ts
DenisPopov15: @inga-affinidi In my comment - there I explicitly stayed about this update only in term of backward compatibility , and have it only while we do not updating this class as abstract (just to support exists usage), and this thing should be not optional when we moving to this class to abstract. I was under the assumption that it was clear, so all this your long comment not really required.

DenisPopov15 (approved):

Yengas (approved):

Volland (approved):

PR #104, opened 2021-06-18
Refactoring for v6 beta

inga-affinidi:
This PR introduces some improvements in maintainability.

(Note that it might be easier to review it commit-by-commit)

# Highlights

## All option objects are now non-optional

SDK no longer assumes that, if you did not pass any `SdkOptions` to it, you must have wanted to use `dev` config. Other services no longer assume that if nobody passed `keyStorageUrl`, you must have wanted to use `stage` service url. No more implicit defaults (except for what was called `setEnvironmentVariables` before, which returned all the service URLs depending on the `dev` parameter).

`CommonNetworkMember.fromSeed(seed)` no longer works; `CommonNetworkMember.fromSeed(seed, { env: 'dev' })` does.

## Refactored remote services usage

All remote services now have their own classes instead of one generic `API`.

And all have their types inferred from swagger docs!

No more
```
keyStorageUrl = keyStorageUrl || STAGING_KEY_STORAGE_URL

const url = `${keyStorageUrl}/api/v1/keys/readMyKey`

const headers = {
  authorization: accessToken,
}

const api = new API(registryUrl, issuerUrl, verifierUrl, options)

const { body } = await api.execute(null, {
  url,
  headers,
  method: 'GET',
})

const { encryptedSeed } = body

return encryptedSeed
```

Instead now we have:
```
const apiService = new KeyStorageApiService(options.keyStorageUrl, options.apiKey)
const { body: { encryptedSeed } } = apiService.readMyKey(accessToken)
```
and it all comes with automatic type annotations, inferred automatically from Swagger docs!

Maybe we could also use that approach in the future to refactor `Affinity` class in `@affinidi/common` so that it will also use `RegistryApiService` instead of doing manual untyped requests.

## Improved type inference for complex CommonNetworkMember inheritance chains
```
class Wallet<TOptions> extends CommonNetworkMember<TOptions> {}
class SuperWallet<TOptions> extends Wallet<TOptions> {}
class HyperWallet extends SuperWallet<HyperOptions> {}

const wallet = HyperWallet.fromSeed(seed, options)
```
In this PR, `wallet` now has the type of `HyperWallet` (not just `CommonNetworkMember`), and `options` are expected to be of `HyperOptions` type (not just `SdkOptions`).

## Reorganized user management code
`CognitoIdentityService` (previously `CognitoService`) now is only a wrapper around `aws-sdk`, with types, convenient methods and error handling.

Actual business logic (e.g. "if someone who already has an unconfirmed account is trying to sign up again, we should remove that account using admin tools and try again"), which was previously located in both `CognitoIdentityService` and `CommonNetworkMember`, now resides in `UserManagementService`, for a clearer separation of boundaries (and for a smaller `CommonNetworkMember` size).

## Reorganized user management code flow
Before:
```
signUpStep1(several parameters) {
  const isUsername = is it an username or a phone number/email?

  if (isUsername) {
    do something
  } else {
    do something else
  }

  signUpStep2(several parameters)

  if (isUsername) {
    do something
  } else {
    do something else
  }
}

signUpStep2(several parameters) {
  const isUsername = is it an username or a phone number/email?

  if (isUsername) {
    do something
  } else {
    do something else
    early return
  }

  signUpStep2(several parameters)

  if (isUsername) {
    do something
  } else {
    do something else
  }
}
...etc
```

After:
```
signUpForEmailOrPhone(several parameters) {
  do something
  do something
  do something
}

signUpForUsername(several parameters) {
  do something
  do something
  do something
}
```

### `CommonNetworkMember` no longer modifies `options` argument
Previously, that caused all kind of bugs (or tests not actually testing what they are supposed to):
```
const options = createGlobalOptionsForLoggedOutUser()

describe(() => {
  it(() => CommonNetworkMember.doSomething(options))

  it(() => CommonNetworkMember.doSomethingElse(options)) // Oops, these options contain `CognitoUserTokens`, because `doSomething` did `options.cognitoUserTokens = tokens` inside
})
```

### `@affinidi/common` no longer has `@affinidi/wallet-core-sdk` as its dev dependency
No more circular dependencies!


sdk/core/src/shared/getOptionsFromEnvironment.ts
```diff
@@ -0,0 +1,116 @@
...
+  STAGING_METRICS_URL,
+  PROD_METRICS_URL,
+} from '../_defaultConfig'
+
+export const getAccessApiKeyFromOptions = (options: SdkOptions) => {
+  if (options.accessApiKey && !options.apiKey) {
+    return options.accessApiKey
+  }
+
+  const apiKey = options.apiKey || 'testApiKey'
```
DenisPopov15, line 45: @inga-affinidi From which part you made this copy-paste ? , should be NO default value here

sdk/core/src/shared/getOptionsFromEnvironment.ts
```diff
@@ -0,0 +1,116 @@
...
+
+  const apiKeyBuffer = KeysService.sha256(Buffer.from(apiKey))
+  return apiKeyBuffer.toString('hex')
+}
+
+function getBasicOptionsFromEnvironment(options: SdkOptions) {
+  const env = options.env || 'staging'
```
DenisPopov15, line 52: @inga-affinidi let also get rid from default option for env parameter and make it required (should be reflected at changelog)

sdk/core/src/shared/getOptionsFromEnvironment.ts
inga-affinidi: @DenisPopov15 , it is a refactored version of `CommonNetworkMember._setAccessApiKey`: https://github.com/affinityproject/affinidi-core-sdk/blob/2d2bc095a161b1e6fa6b489181bbc6c78d849060/sdk/core/src/CommonNetworkMember.ts#L220-L243

sdk/core/src/shared/getOptionsFromEnvironment.ts
inga-affinidi: OK!

sdk/core/src/shared/getOptionsFromEnvironment.ts
DenisPopov15: ok, I see, then its still left there, this thing is obsolete and we removed this default key which was originally done for softer transaction from allow keys at key middleware a while ago and finished that transaction (so this default key not working anymore). So please remove it, and make validation/check, that apiKey or accessApiKey should be required/provided. Thnx

inga-affinidi: OK

common-libs/common/CHANGELOG.md
```diff
@@ -1,3 +1,13 @@
+# 1.9.1 (2020-06-18)
+* Improved type declarations
+* Breaking changes
```
vitalyrudenko-affinidi, line 3: As I understand, those are "internals" that were exposed? It's just doesn't sound right to update patch version with "breaking changes" in changelog :D

sdk/core/src/shared/getOptionsFromEnvironment.ts
Volland: Lets check how we run integration tests . test could be effected

inga-affinidi: All unit and integration tests still work.

sdk/core/src/shared/getOptionsFromEnvironment.ts
inga-affinidi: Fixed

DenisPopov15 (dismissed):

sdk/core/src/services/GenericApiService.ts
```diff
@@ -0,0 +1,117 @@
...
+    options: BasicRequestOptions & { params?: unknown },
+  ) {
+    const { authorization, storageRegion, params, urlPostfix } = options
+    const fetchOptions = {
+      headers: {
+        Accept: 'application/json',
+        'Content-Type': 'application/json',
+        'Api-Key': accessApiKey,
+        ...(authorization && { Authorization: authorization }),
+        ...(storageRegion && { 'X-DST-REGION': storageRegion }),
```
Yengas, line 80: since we created a common api service, would it be a good idea to roll out `x-sdk-version` header with this version?

sdk/browser/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@affinidi/wallet-browser-sdk",
-  "version": "5.0.1",
+  "version": "6.0.0.beta-0",
```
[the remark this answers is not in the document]
inga-affinidi: Not sure what do you mean?

I thought `-beta.0` for the first version was a standard? And the next one will be `-beta.1`.

sdk/core/test/integration/otp/index.test.ts
```diff
@@ -80,7 +80,7 @@ describe('CommonNetworkMember [OTP]', () => {
     commonNetworkMember = new CommonNetworkMember(password, encryptedSeed, options)
 
     await commonNetworkMember.storeEncryptedSeed('', '', accessToken)
-    await commonNetworkMember.signOut()
+    await commonNetworkMember.signOut(options)
```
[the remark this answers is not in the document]
inga-affinidi: Unfortunately it is a legacy code, `commonNetworkMember.signOut` expects `SdkOptions` as a parameter and does not use options that were passed to the constructor before. The only relevant change here is that now options are required (because we no longer use `dev` as a default environment).

Also see https://lemmatree.slack.com/archives/C025J7PEG80/p1624032996226700?thread_ts=1624032991.226600&cid=C025J7PEG80

sdk/core/src/services/GenericApiService.ts
inga-affinidi: I think it would.

Not in this version though; I have certain time constraints with other tasks now.

sdk/core/test/integration/otp/index.test.ts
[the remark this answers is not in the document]
inga-affinidi: I fully agree that it should use class options instead. I wish someone told us why was it made that way. Apparently it was that way from the beginning: https://github.com/affinityproject/affinidi-core-sdk/blob/3a28e957b061dcdaf1e063d7168e5cd0df1e3eec/sdk/core/src/CommonNetworkMember.ts#L641

Yengas (approved):

DenisPopov15 (approved):

PR #109, opened 2021-07-02
[FTL-1705] Create wallet-node-sdk

inga-affinidi:
A quick fix for FTL-1705, wallet-node-sdk is the exact copy of wallet-browser-sdk in this version.
(Will be refactored later in https://github.com/affinityproject/affinidi-core-sdk/pull/108, among with everything else)


vitalyrudenko-affinidi (approved):
lgtm

sdk/node/src/AffinityWallet.ts
```diff
@@ -0,0 +1,87 @@
+import { EventComponent } from '@affinidi/affinity-metrics-lib'
+import { CommonNetworkMember as CoreNetwork, __dangerous } from '@affinidi/wallet-core-sdk'
+
+import platformEncryptionTools from './PlatformEncryptionTools'
+
+export type SdkOptions = __dangerous.SdkOptions & {
+  issueSignupCredential?: boolean
+}
+
+const COMPONENT = EventComponent.AffinidiBrowserSDK
```
Yengas, line 10: I think this is the enum that is used when sending SDK metrics. Did we miss creating a new enum and using it here?

inga-affinidi: This was intended to be a quick fix to unblock work on BBS task. You're right, this enum should get a new value, and I'm going to do this after I'll return from vacation

inga-affinidi: (The problem here is that EventComponent is defined in affinity-metrics-lib which is another project and package entirely)

Yengas (dismissed):

sdk/node/src/AffinityWallet.ts
Yengas: Please don't forget to take a note / create task for it.

vitalyrudenko-affinidi (approved):

Yengas (approved):

PR #114, opened 2021-07-12
FLT-1571 - hmac function for credential ID and types

YuriiSkrypnyk:
Fixes [FLT-1571](https://replika.atlassian.net/browse/FTL-1571) 

Added `signWithPrivateKey` method to `IPlatformEncryptionTools` and added for each implementation.
Added unit tests at each platform.
Upgraded version of `core`, `browser`, `node`, `expo` and `react-native` to `6.0.0-beta.2` version.

### DIFF
--- a/sdk/browser/test/unit/PlatformEncryptionTools.test.ts
+++ b/sdk/browser/test/unit/PlatformEncryptionTools.test.ts
@@ -50,13 +50,23 @@ describe('PlatformEncryptionTools', () => {
     expect(response).to.eql(badEncryptedDataObject)
   })
 
-  it('#signWithPrivateKey', async () => {
+  it('#computePersonalHash', async () => {
     const keysService = new KeysService(encryptedSeed, password)
     const privateKey = keysService.getOwnPrivateKey()
 
-    const hash = await platformEncryptionTools.signWithPrivateKey(privateKey, signedCredential.id)
+    const hash = await platformEncryptionTools.computePersonalHash(privateKey, signedCredential.id)
     const result = await hmacSha256Verify(privateKey, Buffer.from(signedCredential.id), Buffer.from(hash, 'hex'))
 
     expect(result).to.true
+  })
+
+  it('#computePersonalHashWithSameArguments', async () => {
+    const keysService = new KeysService(encryptedSeed, password)
+    const privateKey = keysService.getOwnPrivateKey()
+
+    const firstHash = await platformEncryptionTools.computePersonalHash(privateKey, signedCredential.id)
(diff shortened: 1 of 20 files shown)


sdk/react-native/test/unit/PlatformEncryptionTools.test.ts
```diff
@@ -3,10 +3,12 @@
 import sinon from 'sinon'
 import { expect } from 'chai'
 import { KeysService } from '@affinidi/common'
+import { hmacSha256Verify } from 'eccrypto-js'
```
Volland, line 6: @YuriiSkrypnyk @DenisPopov15 Have we check that eccrypto-js works well on android and ios ?

Yengas (approved):

sdk/react-native/test/unit/PlatformEncryptionTools.test.ts
Yengas: It was already used by the each SDK in `PlatformEncryptionTools.ts` so it should be safe.

sdk/core/src/shared/interfaces.ts
```diff
@@ -6,4 +6,5 @@ export interface IPlatformEncryptionTools {
   platformName: string
   decryptByPrivateKey(privateKeyBuffer: Buffer, encryptedDataString: string): Promise<any>
   encryptByPublicKey(publicKeyBuffer: Buffer, data: unknown): Promise<string>
+  signWithPrivateKey(privateKeyBuffer: Buffer, data: string): Promise<string>
```
inga-affinidi, line 9: What do we need this method for?

If it's needed to transform credential IDs and types for storage, then maybe it would be more appropriate to name it something like `computePersonalHash`.

The way it is named now, it is not clear if two subsequent calls to `signWithPrivateKey` will result in the same value.

DenisPopov15 (approved):

sdk/browser/test/unit/PlatformEncryptionTools.test.ts
```diff
@@ -47,4 +49,14 @@ describe('PlatformEncryptionTools', () => {
 
     expect(response).to.eql(badEncryptedDataObject)
   })
+
+  it('#signWithPrivateKey', async () => {
```
inga-affinidi, line 53: maybe also add an additional test that will check that two calls to `signWithPrivateKey` with the same arguments will result in the same value

Volland (approved):

inga-affinidi (approved):

Yengas (approved):

Volland (approved):

PR #116, opened 2021-07-19
FTL-1574 Implemented AffinidiVaultApiService

inga-affinidi:
* ApiServices moved out to a separate package
  * With TS compiler set to strict mode, for better type safety.
  * And with appropriate changes to the code so that it will build under the strict mode.
  * Plus better support for arrays.
  * Plus better support for some Swagger quirks we didn't encounter before (`allOf`, `description: 'No content'`, missing `additionalProperties` field).
  * Plus support for Swagger enums.
  * Plus support for Swagger query and path parameters (which are now passed as typed objects rather than `urlPostfix` string parameter).
  * In order for ApiServices to work in a separate package, I had to move `SdkError` declaration into `common` package (while leaving the list of error codes in `wallet-core-sdk` as `SdkErrorFromCode`). Both classes now have somewhat stronger types.
  * Another benefit of ApiServices being in a separate package is that in the future we'll be able to get rid of [manual work](https://github.com/affinityproject/affinidi-core-sdk/blob/aeec3f2f5e4c08a29353afa665a10d03df1e6227/common-libs/common/src/Affinity.ts#L50) with API in `common`
* Implemented `AffinidiVaultApiService`
  * Code taken from branch by @YuriiSkrypnyk https://github.com/affinityproject/affinidi-core-sdk/tree/yurii.s/FTL-1574/affinidi-vault-p1.
  * And modified slightly to work with typed query and path parameters.
  * Plus minor `SearchCredentials` bugfix.
  * Swagger spec taken from https://affinidi-vault.dev.affinity-project.org/api/swagger and **not modified**


common-libs/internal-api-clients/src/index.ts
```diff
@@ -0,0 +1,15 @@
...
+import RevocationApiService from './services/RevocationApiService'
+import VerifierApiService from './services/VerifierApiService'
+
+export {
+  BloomVaultApiService,
+  IssuerApiService,
+  KeyStorageApiService,
+  RegistryApiService,
+  RevocationApiService,
+  VerifierApiService,
```
Yengas, line 14: Affinidi Vault is not exported

common-libs/internal-api-clients/src/openapi/_affinidiVault.ts
```diff
@@ -0,0 +1,297 @@
+/* eslint-disable */
```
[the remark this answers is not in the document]
inga-affinidi: You mean generated with an npm script?

Right now everything between `export default` and `as const` is just a copy of https://affinidi-vault.dev.affinity-project.org/api/swagger as-is

sdk/core/src/shared/SdkErrorFromCode.ts
```diff
@@ -1,45 +1,14 @@
...
-  private readonly _originalError: any
-  private readonly _httpStatusCode: any
+import { SdkError } from '@affinidi/common'
 
-  constructor(codeOrError: any, context: any = {}, originalError: any = {}) {
-    const isCode = typeof codeOrError === 'string'
-
-    const code = isCode ? codeOrError : codeOrError.code
-    const error = isCode ? SdkError.errors[code] : codeOrError
+export default class SdkErrorFromCode extends SdkError {
```
vitalyrudenko-affinidi, line 3: What is the context behind this change? I'm a bit hesitant about naming and this kind of nesting in general, looks like OperationError.

common-libs/internal-api-clients/src/services/GenericApiService.ts
```diff
@@ -80,7 +80,7 @@ export default class GenericApiService<TApiSpec extends GenericApiSpec> {
         ...(storageRegion && { 'X-DST-REGION': storageRegion }),
       },
       method,
-      ...(params && { body: JSON.stringify(params, null, 2) }),
+      ...(!!params && { body: JSON.stringify(params, null, 2) }),
```
vitalyrudenko-affinidi, line 83: Does this change anything?

common-libs/internal-api-clients/src/openapi/_affinidiVault.ts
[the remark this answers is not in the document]
YuriiSkrypnyk: yes, so in case something changed in definitions for some service would be nice to keep them in sync at sdk. Just regenerate (get new copies) during build.

common-libs/internal-api-clients/src/services/GenericApiService.ts
inga-affinidi: `internal-api-clients` is compiled in strict TS mode, while `wallet-core-sdk` is compiled in loose TS mode.

`typeof params` is any, and in the original version of the code it was causing compilation error in strict mode because object spreads can only be created from the object types (because if `params` is false-ish, we would end up with `...params`).

sdk/core/src/shared/SdkErrorFromCode.ts
inga-affinidi: `internal-api-clients` throws `SdkError` in some cases, so `SdkError` had to be moved to `common`.

However, specific SDK error codes (`COR-XXX`) clearly do not belong in `common`, so I left this functionality in `wallet-core-sdk`.

common-libs/internal-api-clients/src/openapi/_affinidiVault.ts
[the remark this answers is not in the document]
inga-affinidi: Maybe later, in another task.

common-libs/internal-api-clients/src/index.ts
inga-affinidi: Fixed

YuriiSkrypnyk (approved):
LGTM

vitalyrudenko-affinidi (approved):

PR #117, opened 2021-07-19
[FTL-1755] - affinidi-vault integration (phase #1)

YuriiSkrypnyk:
Added/updated business logic in `WalletStorageService` (as well as `CommonNetworkMember`) to work with affinidi-vault service:
 - `saveCredentials` - hashing credentialId and types to store separately from VC. Removed storing logic to bloom-vault.
 - `getAllCredentials` - get combined list of affinidi-vault and bloom-vault.
 - `getCredentialsByType` - get combined list of affinidi-vault and bloom-vault by types.
 - `getCredentialsByShareToken` - get combined list of affinidi-vault and bloom-vault by credential share request token.
 - `getCredentialById` - get credential by Id from affinidi-vault, in case of error try bloom-vault.
 - `deleteCredentialById` - delete credential by id from affinidi-vault, in case of error try bloom-vault.
 - `deleteAllCredential` - delete all credentials in for loop for affinidi-vault and by index range for bloom-vault.

Created 2 services `AffinidiVaultStorageService` and `BloomVaultStorageService` for separating vaults business logic and make `WalletStorageService` more looking as adapter.

Added integration tests for the use cases above.
Updated version to 6.0.0-beta.4


sdk/core/src/services/WalletStorageService.ts
```diff
@@ -371,15 +365,137 @@ export default class WalletStorageService {
...
+
+  public async deleteAllCredentials(storageRegion?: string): Promise<void> {
+    const accessToken = await this.authorizeVcBloomVault()
+    storageRegion = storageRegion || this._storageRegion
+
+    try {
+      await this._bloomVaultApiService.deleteCredentials({ accessToken, storageRegion, start: 0, end: 99 })
+    } catch (error) {
+      throw new SdkErrorFromCode('COR-0', {}, error)
+    }
```
YuriiSkrypnyk, line 498: Are we going to support `deleteAllCredentials` for affinidi-vault? If that is the case I would prefer to have batch delete on vault's side instead of having individual delete request for each credential here.

common-libs/internal-api-clients/src/services/AffinidiVaultApiService.ts
```diff
@@ -24,10 +24,15 @@ export default class AffinidiVaultApiService extends GenericApiService<ApiSpec>
   }
 
   async searchCredentials(accessToken: string, storageRegion: string, types: string[][]) {
+    const queryParams: any = {}
```
inga-affinidi, line 27: Please avoid using `any` whenever possible, because we lose type checking that way.

One of the alternative approaches:
```
queryParams: (types.length > 0) ? { types: JSON.stringify(types) } : {},
```

sdk/core/src/_defaultConfig.ts
```diff
@@ -6,7 +6,8 @@ export const DEV_REGISTRY_URL = 'https://affinity-registry.dev.affinity-project.
 export const DEV_ISSUER_URL = 'https://affinity-issuer.dev.affinity-project.org'
 export const DEV_VERIFIER_URL = 'https://affinity-verifier.dev.affinity-project.org'
 export const DEV_KEY_STORAGE_URL = 'https://affinity-wallet-backend.dev.affinity-project.org'
-export const DEV_VAULT_URL = 'https://bloom-vault.dev.affinity-project.org'
+export const DEV_BLOOM_VAULT_URL = 'https://bloom-vault.dev.affinity-project.org'
+export const DEV_VAULT_URL = 'https://affinidi-vault.dev.affinity-project.org'
```
inga-affinidi, line 10: Better rename it to `DEV_AFFINIDI_VAULT_URL` so that it won't be misleading (so that developers who relied on `DEV_VAULT_URL` will at least learn about backwards incompatibility straight away at compile time rather than silently getting the wrong URL)

sdk/core/src/services/WalletStorageService.ts
```diff
@@ -51,18 +54,23 @@ type ConstructorOptions = {
...
     keysService: KeysService,
+    didAuthService: DidAuthService,
     platformEncryptionTools: IPlatformEncryptionTools,
     options: ConstructorOptions,
   ) {
     this._keysService = keysService
+    this._didAuthService = didAuthService
     this._platformEncryptionTools = platformEncryptionTools
-    this._bloomVaultApiService = new BloomVaultApiService(options)
+    this._bloomVaultApiService = new BloomVaultApiService({ ...options, vaultUrl: options.bloomVaultUrl })
```
inga-affinidi, line 71: Or better to do it explicitly now, when there are several vault urls: `new BloomVaultApiService({ accessApiKey: options.accessApiKey, vaultUrl: options.bloomVaultUrl })`

sdk/core/src/services/WalletStorageService.ts
```diff
@@ -340,18 +298,54 @@ export default class WalletStorageService {
     return credentials
   }
 
-  async getCredentialByIndex(credentialIndex: number): Promise<any> {
-    const paginationOptions: FetchCredentialsPaginationOptions = { skip: credentialIndex, limit: 1 }
-    const credentials = await this.fetchDecryptedCredentials(paginationOptions)
+  private async _authorizeVcVault(audienceDid: string): Promise<string> {
```
inga-affinidi, line 301: Please rename to `authorizeAffinidiVcVault` so that it is clear what methods are for legacy bloom vault and what are for new affinidi vault

sdk/core/src/services/WalletStorageService.ts
```diff
@@ -371,15 +365,137 @@ export default class WalletStorageService {
     throw new SdkErrorFromCode('COR-23', { id })
   }
 
-  async deleteCredential(id?: string, credentialIndex?: number): Promise<void> {
-    if ((credentialIndex !== undefined && id) || (!id && credentialIndex === undefined)) {
-      throw new SdkErrorFromCode('COR-1', {
-        errors: [{ message: 'should pass either id or credentialIndex and not both at the same time' }],
+  public async saveCredentials(audienceDid: string, credentials: SignedCredential[], storageRegion?: string) {
```
inga-affinidi, line 368: So `WalletStorageService` gets some of the auth info as constructor params (as two params, `keysService` and `didAuthService` with somewhat duplicated data, and some as method params, on every method call.

But `audienceDid` should not change between different calls to the same instance of `WalletStorageService`. So why not to pass it to the constructor as well, at least?

sdk/core/src/services/WalletStorageService.ts
```diff
@@ -371,15 +365,137 @@ export default class WalletStorageService {
...
+    try {
+      const { body, status } = await this._vaultApiService.getCredential(token, storageRegion, hashedId)
+      if (status >= 400) {
+        throw new SdkErrorFromCode('COR-23', { credentialId })
+      }
+
+      encryptedCredential = body.payload
+    } catch (error) {
+      // should be deleted during migration to affinidi-vault Phase #2
+      const bloomToken = await this.authorizeVcBloomVault()
```
inga-affinidi, line 444: We probably should only do that for 404 error, not just for any exception that was thrown from `getCredential`.

sdk/core/src/services/WalletStorageService.ts
```diff
@@ -371,15 +365,137 @@ export default class WalletStorageService {
...
+      const credentialIndexToDelete = await this._findCredentialIndexById(credentialId)
```
inga-affinidi, line 445: It's not `ToDelete` :)

sdk/core/src/services/WalletStorageService.ts
```diff
@@ -371,15 +365,137 @@ export default class WalletStorageService {
...
+      const { body, status } = await this._bloomVaultApiService.getCredentials({
+        accessToken: bloomToken,
+        start: credentialIndexToDelete,
+        end: credentialIndexToDelete,
+        storageRegion,
       })
+      if (status >= 400 && body?.length > 0) {
```
inga-affinidi, line 452: `GenericApiService` throws an error for any status that does not start with `2`. So this condition will never be satisfied (and so is the one above for `AffinidiVault`).

sdk/core/src/services/WalletStorageService.ts
```diff
@@ -371,15 +365,137 @@ export default class WalletStorageService {
...
+  }
+
+  public async getCredentialById(audienceDid: string, credentialId: string, storageRegion?: string): Promise<any> {
+    const token = await this._authorizeVcVault(audienceDid)
+    storageRegion = storageRegion || this._storageRegion
+    const privateKeyBuffer = this._keysService.getOwnPrivateKey()
+
+    const hashedId = await this._platformEncryptionTools.computePersonalHash(privateKeyBuffer, credentialId)
+
+    let encryptedCredential: string
```
inga-affinidi, line 434: Extracting this into a separate (private) method `getEncryptedCredentialById` might make code simpler.

(Try to get it from Affinidi Vault, return it; if there was an error: throw it unless it was `AVT-2`, in which case try to get it from Bloom Vault and return it.)

sdk/core/src/services/WalletStorageService.ts
```diff
@@ -371,15 +365,137 @@ export default class WalletStorageService {
...
+    return credential
+  }
+
+  public async deleteCredentialById(audienceDid: string, credentialId: string, storageRegion?: string) {
+    const token = await this._authorizeVcVault(audienceDid)
+    storageRegion = storageRegion || this._storageRegion
+
+    try {
+      const { status } = await this._vaultApiService.deleteCredential(token, storageRegion, credentialId)
+      if (status >= 400) {
```
inga-affinidi, line 470: This condition is always false.

sdk/core/src/services/WalletStorageService.ts
inga-affinidi: There won't be batch delete on vault side, because it is impractical or impossible to do in a single transaction, especially for multi-region storage. Relevant discussion: https://lemmatree.slack.com/archives/C025J7PEG80/p1623679947068300?thread_ts=1623673672.040700&cid=C025J7PEG80

Handling of errors related to the incomplete delete should be done on a client side.
(And also turns out there was a bug in original `WalletStorageService` which remained in the updated version; it only deletes first 100 credentials from bloom vault.)

sdk/core/src/shared/getOptionsFromEnvironment.ts
```diff
@@ -78,6 +81,7 @@ function getBasicOptionsFromEnvironment(options: EnvironmentOptions) {
         issuerUrl: options.issuerUrl || DEV_ISSUER_URL,
         registryUrl: options.registryUrl || DEV_REGISTRY_URL,
         verifierUrl: options.verifierUrl || DEV_VERIFIER_URL,
+        bloomVaultUrl: options.vaultUrl || DEV_BLOOM_VAULT_URL,
         vaultUrl: options.vaultUrl || DEV_VAULT_URL,
```
inga-affinidi, line 85: This is certainly a bug: `options.vaultUrl` is used here both for bloom vault and for affinidi vault.

We should probably replace `vaultUrl` in `EnvironmentOptions` with `affinidiVaultUrl` and `bloomVaultUrl`.

sdk/core/test/helpers/testPlatformTools.ts
```diff
@@ -2,13 +2,17 @@ import { IPlatformEncryptionTools } from '../../src/shared/interfaces'
 
 export const testPlatformTools: IPlatformEncryptionTools = {
   platformName: 'stub',
-  decryptByPrivateKey: async () => {
-    throw new Error('not implemented')
+  // eslint-disable-next-line no-unused-vars
+  decryptByPrivateKey: async (privateKeyBuffer: Buffer, data: string) => {
+    return data
```
inga-affinidi, line 7: `decryptByPrivateKey` and `encryptByPublicKey` should be complementary operations.
So we'll need to do `JSON.parse` here.

sdk/core/test/integration/NetworkMember.test.ts
```diff
@@ -1047,4 +1048,45 @@ describe('CommonNetworkMember', () => {
...
+  it('#getAllCredentials by types', async () => {
+    const commonNetworkMember = new CommonNetworkMember(password, encryptedSeedElem, options)
+
+    const result = await commonNetworkMember.getAllCredentials([credential.type])
+
+    const parsedCredential = JSON.parse(result[0])
+    expect(parsedCredential.id).to.eql(credential.id)
+  })
+
+  it('#deleteCredentialById', async () => {
```
inga-affinidi, line 1088: All these tests only check that `CommonNetworkMember` works with the new vault.

They do not check that `CommonNetworkMember` works with both vaults (that it returns credentials from both vaults, that it looks for a credential with requested ID in bloom vault if not found in affinidi vault; that it deletes credential from bloom vault if not found in affinidi vault)

sdk/core/src/services/WalletStorageService.ts
```diff
@@ -304,18 +274,6 @@ export default class WalletStorageService {
...
-
-    for (const item of data) {
-      const cyphertext = await this._platformEncryptionTools.encryptByPublicKey(publicKeyBuffer, item)
-      encryptedCredentials.push(cyphertext)
-    }
-
-    return await this.saveCredentials(encryptedCredentials, storageRegion)
-  }
-
   async fetchAllDecryptedCredentials() {
```
inga-affinidi, line 277: This method is used in `AffinityWallet.getCredentials` method in browser, expo, node, and react-native SDKs, and this is what API users actually use.

This means that all the API users, even after they upgrade to new SDK and fix all incompatibilities, will still only be reading from the old SDK, unless they replace calls to `AffinityWallet.getCredentials` with `AffinityWallet.getAllCredentials` (which does not even support handling `credentialShareRequestToken`.

One of the solutions would be to update all the specific `getCredentials` methods to call `WalletStorageService.getAllCredentials` instead of `fetchAllDecryptedCredentials`.

Then there is a problem with expo and react-native caching all fetched credentials in AffinityWallet object; I'll start a Slack discussion on that topic now.

Also, please make sure that all methods that only work with bloom vault are renamed to make this clear (to avoid confusion). For example, this method could be renamed to `fetchAllDecryptedCredentialsFromBloomVault`.

inga-affinidi (changes requested):

sdk/core/src/CommonNetworkMember.ts
```diff
@@ -176,7 +178,9 @@ export abstract class CommonNetworkMember<TOptions extends SdkOptions = SdkOptio
     this._revocationApiService = new RevocationApiService({ revocationUrl, accessApiKey })
     this._userManagementService = new UserManagementService({ clientId, userPoolId, keyStorageUrl, accessApiKey })
     this._keyManagementService = new KeyManagementService({ keyStorageUrl, accessApiKey })
-    this._walletStorageService = new WalletStorageService(keysService, platformEncryptionTools, {
+    const didAuthService = new DidAuthService({ encryptedSeed, encryptionKey: password })
+    this._walletStorageService = new WalletStorageService(keysService, didAuthService, platformEncryptionTools, {
```
Yengas, line 182: can we pass the did to **walletStorageService** here to prevent passing it with each method call?

sdk/core/src/services/WalletStorageService.ts
```diff
@@ -1,6 +1,7 @@
 import { toRpcSig, ecsign } from 'ethereumjs-util'
```
Yengas, line 1: we can convert this service into 3 different classes:
- BloomVaultService: old bloom vault related auth and crud code
- AffinidiVaultService: new affinidi vault related auth, hashing, and crud code
- WalletStorageService: combine BloomVault + AffinidiVault

that should make the code more readable, make it easier to unit test, review 3 different classes.

inga-affinidi:
Also, if you have time for that, you might want to consider making the code simpler an more maintainable, and this PR easier to review, by, for example:

* Renaming the old `WalletStorageService` to `BloomVaultStorageService`
  * Removing everything related to saving credentials
  * Removing relevant parts related to decrypting credentials
  * Basically, it only needs to expose the following public methods: `deleteAllCredentials`, `fetchAllBlobs`, `deleteCredential`. You may also preserve methods like `getSignedCredentials` and `getCredentialOffer` (they do not belong to `WalletStorageService` at all, but that's another task). It might also make sense to implement simple `getCredentialById` there; and since it already has `filterCredentials`, we might as well implement `fetchAllCredentialsByTypes` (if so, we should probably export `fetchAllDecryptedCredentials` instead of `fetchAllBlobs`)
  * So it will be a wrapper around `BloomVaultApiService`, providing somewhat reasonable interface for bloom vault unreasonable API (it will contain all the logic of fetching credentials in batches, etc)
* Creating the new `WalletStorageService` which will work with `AffinidiVaultApiService`, and fall back to `BloomVaultStorageService` as needed; but will not contain any of the complex logic required by bloom vault (fetching credentials in batches, dealing with indices, etc).

That way, when the day comes for us to drop bloom vault support, it will be enough to just remove all references to `BloomVaultStorageService` from `WalletStorageService`. And in the meantime, we won't have all of that legacy code in `WalletStorageService`, making it easier to maintain.

sdk/core/src/services/WalletStorageService.ts
YuriiSkrypnyk: okay, I see problem with batch delete described in that thread and here is method for delete all credentials, which means we don't want to pass list of items to delete, but want delete them all (in all regions), isn't it? We can remove this logic at all if we don't need it or rename it appropriately like `deleteNCredentials` or so?

sdk/core/src/_defaultConfig.ts
YuriiSkrypnyk: My thoughts was about that bloom-vault becomes deprecated and only affinidi-vault left as single credential storage, but yeah that could be renamed until bloom-vault will be removed to avoid issues you described.

sdk/core/src/services/WalletStorageService.ts
YuriiSkrypnyk: I see, `did` created once during first get. Good point!

sdk/core/src/services/WalletStorageService.ts
YuriiSkrypnyk: > get by credential id: should go to both Bloom Vault and Affinidi Vault returning the first successful response. If both fails, return Affinidi Vault error.

YuriiSkrypnyk: that what we have in ADR, so probably make sense do that for every

inga-affinidi: Yes, it looks like the wording in ADR was imprecise.

What it is supposed to mean is that we should try to find a credential with this ID in Affinidi Vault; and go to the Bloom Vault if it is not in the Affinidi Vault. Not that we should try to find it in Bloom Vault if Affinidi Vault returned 500 Internal Server Error, that would make little sense.

The reason for using both Affinidi Vault and Bloom Vault is that some credentials might be stored in Affinidi Vault only, while some others might be stored in Bloom Vault only; and if credential with that id is stored _anywhere_, we should return it no matter where it is stored. Not that Bloom Vault is some kind of a fallback for when Affinidi Vault is unavailable.

YuriiSkrypnyk: Okay, but `500 Internal Server Error` in Affinidi Vault doesn't guarantee that credential exists in Affinidi Vault and not in Bloom Vault. So in worst case we will get 500 at affinidi-vault call and 404 (or some other) in bloom-vault, then will throw error from affinidi-vault. But in case credential present in bloom-vault we reject getting it because affinidi-vault is unavailable.

sdk/core/src/services/WalletStorageService.ts
Yengas: @inga-affinidi regarding [your comment](https://github.com/affinityproject/affinidi-core-sdk/pull/117#issuecomment-883229623), I think we are talking about similar thing. But I think separating to 3 classes could be better. Because both Affinidi and Bloom has different characteristics with their own auth logic and Affinidi Vault having hashing logic. 

WalletStorageService could be the adapter that uses these two services to implement reading from both, merging the results etc. kind of logic.

inga-affinidi: @Yengas  Agree, that's also an option.

sdk/core/src/services/WalletStorageService.ts
inga-affinidi: That's an overcomplication, Bloom Vault is not intended to be a failsafe for Affinidi Vault.

If Affinidi Vault returned 500 Internal Server Error, the best thing to do is to rethrow it to the caller instead of doing any more processing.

Affinidi Vault should not return 500; and if it does, we don't have to write complex logic to handle it in some way and to fallback to Bloom Vault.

YuriiSkrypnyk: Okay, I can handle 404 error only, but all this logic complex left as it, nothing else should be written, except new line:
```
} catch (error) {
      if (error.httpStatusCode === 404) {
```
We still need try/catch as GenericApiService throws an error if status code not starting with `2`, right?

sdk/core/src/services/WalletStorageService.ts
YuriiSkrypnyk: hm, that means we should put calls to bloom-vault into try catch as well, otherwise we will be throwing error from bloom-vault in case both encountered an error

sdk/core/src/services/WalletStorageService.ts
inga-affinidi: Yes, we still need try/catch there.

And also it's probably better to rely on error code than on HTTP status code.

So it could look like:

```
try {
  const { body } = loadFromAffinidiVault(...)
  encryptedCredential = body.payload
} catch (err) {
  if (err.code !== 'AVT-2') {
    throw err
  }

  const blobInfo = loadFromBloomVault(...)
  encryptedCredential = blobInfo.cyphertext
}
```

Or even better, if you separate this into three classes as @Yengas suggested, the entire method could look like:

```
try {
  return loadFromAffinidiVault(id)
} catch (err) {
  if (err.code !== 'AVT-2') {
    throw err
  }

  return loadFromBloomVault(id)
}
```

sdk/core/src/services/WalletStorageService.ts
YuriiSkrypnyk: After combining all the logic in WalletStorageService from both vaults looks like will be better to separate them out. I thought about it initially, but was hesitating that we will have pretty long dependency chain of services.

sdk/core/src/services/WalletStorageService.ts
YuriiSkrypnyk: @DenisPopov15 @Yengas do you have some thought here?

sdk/core/src/services/WalletStorageService.ts
YuriiSkrypnyk: yes, that what I just did. Currently working on splitting logic to separate services

sdk/core/src/services/WalletStorageService.ts
inga-affinidi: This is still not addressed.

Please implement FTL-1755 as described in earlier comments:
* `AffinidiVaultStorageService` as a wrapper over `AffinidiVaultApiService` with all the required authorization / encryption / decryption / hashing logic (you already did this part);
* `BloomVaultStorageService` as a wrapper over `BloomVaultApiService` (it will basically be a copy of `WalletStorageService` prior to this PR)
* `WalletStorageService` with interface similar to `AffinidiVaultStorageService`, implementing the logic of "fetch single credential from AffinidiVault, if not found there, fetch it from BloomVault", "delete single credential from AffinidiVault, if not found there, delete it from BloomVault" and "fetch credentials by types from AffinidiVault, concatenate with filtered credentials from BloomVault".

And please make sure that everything builds and lint checks are passing successfully and all tests are passing successfully for dependent projects (`browser`, `expo`, `node`, `react-native`) as well.

YuriiSkrypnyk: Yes, that is not finished yet, forgot move it to draft.

sdk/browser/test/unit/AffinityWallet.test.ts
```diff
@@ -104,64 +86,21 @@ describe('AffinityWallet', () => {
...
-
-    it('should remove with given index', async () => {
-      const credentialIndex = 1
-
-      sinon.stub(CommonNetworkMember.prototype as any, 'deleteCredentialByIndex').resolves(undefined)
-
-      const affinityWallet = new AffinityWallet(walletPassword, encryptedSeed, sdkOptions)
-
-      const result = await affinityWallet.deleteCredential(undefined, credentialIndex)
+      const result = await affinityWallet.deleteCredentialById(signedCredential.id)
```
inga-affinidi, line 94: This test does not test anything now.

Maybe we should mock `WalletStorageService.deleteCredentialById` instead.

(But overall this is yet another example of how useless our unit tests are...)

sdk/core/test/unit/services/WalletStorageService.test.ts
```diff
@@ -99,277 +97,159 @@ describe('WalletStorageService', () => {
...
+    await walletStorageService.deleteCredentialById(signedCredential.id)
+  })
 
-      {
-        const expectedResult = Array(50).fill(signedCredential)
-        expectedResult[30] = { cyphertext: null }
-        const expectedPath = `${fetchCredentialsBase}200/299`
-        nock(STAGING_AFFINIDI_VAULT_URL).get(expectedPath).reply(200, expectedResult)
-      }
+  it('#getCredentialsById not found in Affinidi vault and checking in Bloom vault', async () => {
```
inga-affinidi, line 221: "deleteCredentialsById"

sdk/core/test/unit/services/WalletStorageService.test.ts
```diff
@@ -99,277 +97,159 @@ describe('WalletStorageService', () => {
...
-      { args: [{ skip: 1, limit: 1 }], path: '1/1' },
-    ]
-
-    for (const { args, path } of cases) {
-      it(`should request correct path with args ${JSON.stringify(args)}`, async () => {
-        await authorizeVault()
-
-        const expectedPath = fetchCredentialsBase + path
-        nock(STAGING_AFFINIDI_VAULT_URL).get(expectedPath).reply(200, [signedCredential])
+  it('#getCredentialsById not found', async () => {
```
inga-affinidi, line 236: "deleteCredentialById"

sdk/core/test/unit/services/WalletStorageService.test.ts
```diff
@@ -99,277 +97,159 @@ describe('WalletStorageService', () => {
...
+      .resolves()
 
-      const walletStorageService = createWalletStorageService()
-      const allBlobs = await walletStorageService.fetchAllBlobs()
-
-      expect(allBlobs.length).to.eql(247)
-      expect(allBlobs).to.eql(Array(247).fill(signedCredential))
-    })
+    const walletStorageService = createWalletStorageService()
+    await walletStorageService.deleteCredentialById(signedCredential.id)
```
inga-affinidi, line 233: should probably also check here that the second stub is called at least

sdk/core/src/services/BloomVaultStorageService.ts
```diff
@@ -0,0 +1,293 @@
...
+      if (error.httpStatusCode === 404) {
+        throw new SdkErrorFromCode('COR-14', {}, error)
+      } else {
+        throw error
+      }
+    }
+  }
+
+  /* istanbul ignore next: private function */
+  private async _fetchEncryptedCredentials(
```
inga-affinidi, line 147: This method does not seem to be used anywhere?

I wonder why IDE did not highlight it as unused (since it's private)

sdk/core/src/services/WalletStorageService.ts
```diff
@@ -293,93 +94,100 @@ export default class WalletStorageService {
...
-    }
+    const responses = await this._affinidiVaultStorageService.saveCredentials(credentials, storageRegion)
+
+    return responses
   }
 
-  async encryptAndSaveCredentials(data: unknown[], storageRegion?: string) {
-    const publicKeyBuffer = this._keysService.getOwnPublicKey()
-    const encryptedCredentials = []
+  public async getAllCredentials(types?: string[][], storageRegion?: string) {
```
inga-affinidi, line 105: But it's not "all", it's more like `getCredentialsByTypes` (similar to `getCredentialsByToken`).

I'd suggest to have three methods:
* `getAllCredentials(storageRegion)`
* `getCredentialsByTypes(types, storageRegion)`, with required types parameter
* `getCredentialsByToken(token, storageRegion)`, with required token parameter (that is, without handling a special case of token being empty)

sdk/core/src/services/WalletStorageService.ts
```diff
@@ -293,93 +94,100 @@ export default class WalletStorageService {
...
+    } catch (error) {
+      if (error.code !== 'AVT-2') {
+        throw error
+      }
 
-    return credentials[0]
+      // should be deleted during migration to affinidi-vault Phase #2
+      try {
+        return await this._bloomVaultStorageService.getCredentialById(credentialId, storageRegion)
+      } catch (bloomError) {
```
inga-affinidi, line 152: No need to catch anything there.
(And if Bloom Vault returned some unexpected error, probably better if the caller receives it, instead of it being hidden by AVT-2)

sdk/core/src/services/WalletStorageService.ts
```diff
@@ -293,93 +94,100 @@ export default class WalletStorageService {
...
+      if (error.code !== 'AVT-2') {
+        throw error
       }
 
-      if (credentialId && credentialId === id) {
-        return blob.id
+      // should be deleted during migration to affinidi-vault Phase #2
+      try {
+        await this._bloomVaultStorageService.deleteCredentialById(storageRegion, credentialId)
+      } catch (bloomError) {
```
inga-affinidi, line 171: Same as above, no need for try..catch block here

sdk/core/src/shared/SdkErrorFromCode.ts
```diff
@@ -146,6 +146,11 @@ export default class SdkErrorFromCode extends SdkError {
         message: 'did method: {{didMethod}} not supported. Supported methods: {{supportedDidMethods}}',
         httpStatusCode: 409,
       },
+      'COR-26': {
```
inga-affinidi, line 149: But it is not thrown anywhere?

sdk/core/src/CommonNetworkMember.ts
```diff
@@ -1992,20 +1982,49 @@ export abstract class CommonNetworkMember<TOptions extends SdkOptions = SdkOptio
...
   /**
-   * @description Delete credential by id if found in given range
-   * @param id - id of the credential
-   * @param credentialIndex - credential to remove
+   * @description Retrieve all credential
+   * @param types (string[][]) - (optional) specify credential types to filter
+   * @param storageRegion (string) - (optional) specify region where credentials will be stored
+   * @returns a single VC
+   */
+  async getAllCredentials(types?: string[][], storageRegion?: string): Promise<any> {
```
inga-affinidi, line 2000: Same suggestion here, `getAllCredentials` should not have `types` as its parameter.
If we need a method that accepts types (do we need it?), it should be called `getCredentialsByTypes`, while `getAllCredentials` should return _all_ credentials.

inga-affinidi:
Overall it seems good now (besides these comments), good job!

By the way if you need to catch lint errors locally, `npm run lint` should do the job.

sdk/browser/test/unit/AffinityWallet.test.ts
YuriiSkrypnyk: yes, many tests at platforms were testing code from mostly core while core didn't. Make sense to rethink way tests are testing. But this better to delete as for me

sdk/core/src/services/WalletStorageService.ts
YuriiSkrypnyk: So, here I just was following the ADR, which says:

> delete operations: should go to both Bloom Vault and Affinidi Vault. Returning the first successful operation or error of Affinidi Vault in case both fail.

I thought Bloom vault is working as reserve, so main vault provider is Affinidi vault?

sdk/core/test/unit/services/WalletStorageService.test.ts
YuriiSkrypnyk: yes, good point. Ideally we should check all the stubs are getting called, but should check for cases when nothing to check for sure.

sdk/core/src/services/BloomVaultStorageService.ts
YuriiSkrypnyk: it was highlighted, I just missed it, thank you

sdk/core/src/services/WalletStorageService.ts
YuriiSkrypnyk: Removed catch from Bloom Vault call

sdk/core/src/shared/SdkErrorFromCode.ts
YuriiSkrypnyk: I was added this initially, but don't want overwrite existing errors. So will remove it.

sdk/core/src/services/WalletStorageService.ts
YuriiSkrypnyk: I was trying make it, to be similar to Affinidi vault api, but more logically would be separate them for single responsibility.

sdk/core/src/services/AffinidiVaultStorageService.ts
```diff
@@ -0,0 +1,184 @@
...
+    for (const credential of credentials) {
+      let credentialId = credential?.id
+      if (!isW3cCredential(credential)) {
+        credentialId = credential?.data?.id
+      }
+
+      const credentialIdHash = await this._platformEncryptionTools.computePersonalHash(privateKeyBuffer, credentialId)
+
+      const typeHashes = []
+      if (isW3cCredential(credential)) {
```
Yengas, line 56: I am asking because I don't know. What types of credentials do we have? As far as I understand;
- if credential is not w3c, we get credentialId from the the `credential.data.id`
- if credential is not w3c, it does not have a type array
what is the other type of credentials?

sdk/core/src/services/AffinidiVaultStorageService.ts
```diff
@@ -0,0 +1,184 @@
...
+
+    return credential
+  }
+
+  public async saveCredentials(credentials: any[], storageRegion: string): Promise<VaultCredential[]> {
+    const token = await this._authorizeVcVault()
+
+    const responses = []
+
+    const encryptedCredentials = await this._encryptCredentials(credentials)
```
Yengas, line 119: the did auth has some expiration time set by the server if I remember correctly. for this exact reason, I remember setting up token expiration checking / re-auth logic in some frontend. please see: https://gitlab.com/affinidi/foundational/affinity-verifier-portal/-/blob/main/src/services/MessageService.ts#L53

Here after the did auth we are doing N encryption and HTTP requests to the backend with the same token. According to the clients computing power, network speed that expiration can happen. Causing some operations to fail. For that reason; I think it would be better to implement something like the link I've sent, and calling the did-auth everytime before making a request to the server.

sdk/core/src/services/AffinidiVaultStorageService.ts
```diff
@@ -0,0 +1,184 @@
...
+      responses.push(body)
+    }
+
+    return responses
+  }
+
+  public async searchCredentials(types: string[][], storageRegion: string): Promise<any[]> {
+    const token = await this._authorizeVcVault()
+
+    const hashedTypes = await this._computeTypesHashes(types)
```
Yengas, line 136: This operation is probably pretty safe, but looking at this pattern; I think we can create some class in internal api package that handles did-auth. And that helper class can do did auth before each request. For example;

```
export default class AffinidiVaultApiService extends DidAuthApiService<ApiSpec> {
  someMethodWithDidAuth() {
    const token = this.getToken();
    return this.httpRequest(...) 
  }

  someMethod() {
    return this.httpRequest(...)
   }
}

// Here we have the token impl with expiration date check
export default abstract class DidAuthService<T> extends GenericApiService<T> 
```

Yengas (changes requested):

sdk/core/src/services/AffinidiVaultStorageService.ts
inga-affinidi: @Yengas, we definitely should do something about that, but it's a known old problem and it's beyond the scope of this PR.

We use `isW3cCredential` throughout all SDK and cloud-wallet-api, with the assumption that it's either W3c credential (with id and types), or OADocument (with id stored in `.data.id` and without types)

sdk/core/src/services/AffinidiVaultStorageService.ts
inga-affinidi: @Yengas, the same logic was used in bloom vault.

The expiration time is AFAIK not measured in seconds, and if this method takes so long to complete that token expires, we'll probably have some other problem.

sdk/core/src/services/AffinidiVaultStorageService.ts
inga-affinidi: Yes, something like that would be great.

sdk/core/src/CommonNetworkMember.ts
```diff
@@ -1992,13 +1992,22 @@ export abstract class CommonNetworkMember<TOptions extends SdkOptions = SdkOptio
...
+    return this._walletStorageService.getAllCredentials(storageRegion)
+  }
+
+  /**
+   * @description Retrieve all credential by types
+   * @param types (string[][]) - specify credential types to filter
+   * @param storageRegion (string) - (optional) specify region where credentials will be stored
+   * @returns a single VC
+   */
+  async getCredentialsByTypes(types: string[][], storageRegion?: string): Promise<any> {
```
inga-affinidi, line 2009: Again, do we really need this method?

sdk/core/src/services/WalletStorageService.ts
```diff
@@ -102,14 +102,24 @@ export default class WalletStorageService {
...
+    const credentials = await this._affinidiVaultStorageService.searchCredentials([], storageRegion)
 
     // should be deleted during migration to affinidi-vault Phase #2
-    const bloomCredentials = await this._bloomVaultStorageService.getAllCredentials(types, storageRegion)
+    const bloomCredentials = await this._bloomVaultStorageService.searchCredentials([], storageRegion)
+
+    return [...credentials, ...bloomCredentials]
+  }
+
+  public async getCredentialsByTypes(types: string[][], storageRegion?: string) {
```
inga-affinidi, line 116: Maybe we should make this one private and remove `getCredentialsByTypes` from `CommonNetworkMember`.

(The problem with making more methods public is that even though they might be completely useless, we'll still have to support them and maintain backwards compatibility in case one of the external clients relies on them. So it's better to keep the API surface area as small as possible)

sdk/core/test/unit/services/AffinidiVaultStorageService.test.ts
```diff
@@ -0,0 +1,237 @@
...
+
+const mockDidAuth = () => {
+  nock(STAGING_AFFINIDI_VAULT_URL)
+    .post('/api/v1/did-auth/create-did-auth-request')
+    .reply(
+      200,
+      '"eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiIxMjM0NTY3ODkwIiwibmFtZSI6IkpvaG4gRG9lIiwiaWF0IjoxNTE2MjM5MDIyLCJpc3MiOiJkaWQ6ZWxlbTpFaUNILXh4Y25rZ1p2NlF2anZvX1VYbi04RFVkVU4zRXRCSnhvbEFRYlFyQ2NBIyJ9.PC4hlTYT5oc1rtcE3Ngq1LN35vAQXI0QgC2nzzQ9RKw"',
+    )
+
+  nock(STAGING_REGISTRY_URL).post('/api/v1/did-auth/create-did-auth-response').reply(200, {})
```
Yengas, line 40: what endpoint is this? the did auth response was generated on the client side, without any request to the server. I remeber we should only make request to server for getting the did auth request token.

sdk/core/src/services/AffinidiVaultStorageService.ts
YuriiSkrypnyk: That actually reasonable as our sdk allow to store thousands and even more credentials at same time and we cannot just reject this. What we can do that combine token refresh logic with idea  described https://github.com/affinityproject/affinidi-core-sdk/pull/117#discussion_r674595396 and then we actually can reuse token for same `audienceDid` requests in short period of time by keeping it in memory and refresh when it got expired.

YuriiSkrypnyk: But that is going to be as enhancement as for me. Just don't want increase size of current PR

sdk/core/src/services/WalletStorageService.ts
```diff
@@ -293,93 +94,101 @@ export default class WalletStorageService {
...
-    const credentials = []
-    for (const blob of blobs) {
-      const credential = await this._platformEncryptionTools.decryptByPrivateKey(privateKeyBuffer, blob.cyphertext)
-      credentials.push(credential)
-    }
+    // should be deleted during migration to affinidi-vault Phase #2
+    const bloomCredentials = await this._bloomVaultStorageService.searchCredentials(types, storageRegion)
 
-    return credentials
+    return [...credentials, ...bloomCredentials]
```
Yengas, line 124: ADR mentions merging the both results using a `Set<string>` instead of array concat to remove duplicate records.

sdk/core/src/services/AffinidiVaultStorageService.ts
Yengas: Sure I agree it is not exactly related to this PR. But it is maybe a blocker to release v6(I need to check expiration date).

I created [FTL-1770](https://replika.atlassian.net/browse/FTL-1770), I will try to find expiration date and add more info do that task regarding optimizations. If it is a blocker, we should do it before removing the beta suffix. Thanks guys.

sdk/core/src/services/AffinidiVaultStorageService.ts
```diff
@@ -0,0 +1,184 @@
...
+        payload: credential.payload,
+      })
+
+      responses.push(body)
+    }
+
+    return responses
+  }
+
+  public async searchCredentials(types: string[][], storageRegion: string): Promise<any[]> {
```
Yengas, line 133: sending an empty array to affinidi vault will cause it to return nothing as far as I remember that is how it was designed. you either need to send no `?types` parameter. or you need to supply a non empty array.

sdk/core/src/services/AffinidiVaultStorageService.ts
```diff
@@ -0,0 +1,184 @@
...
+    const hashedId = await this._platformEncryptionTools.computePersonalHash(privateKeyBuffer, credentialId)
+
+    await this._vaultApiService.deleteCredential(token, storageRegion, hashedId)
+  }
+
+  public async deleteAllCredentials(storageRegion: string): Promise<void> {
+    const token = await this._authorizeVcVault()
+    const privateKeyBuffer = this._keysService.getOwnPrivateKey()
+
+    const credentials = await this.searchCredentials([], storageRegion)
```
Yengas, line 169: empty array may cause affinidi vault to return nothing

Yengas (changes requested):

sdk/core/src/services/AffinidiVaultStorageService.ts
inga-affinidi: I wonder why tests pass then.

This one, for example: https://github.com/affinityproject/affinidi-core-sdk/blob/yurii.s/FTL-1755/affinidi-vault-p1/sdk/node/test/integration/otp/index.test.ts#L100

inga-affinidi: But overall I agree, empty array means result should be empty.

So even if it works now, we should make the code more clear, and pass `undefined` to this method when we want to get all credentials without filtering.

(Or `[[]]`, but IIRC affinidi vault explicitly disallows that)

YuriiSkrypnyk: hm, okay I didn't know that and that is why put this check https://github.com/affinityproject/affinidi-core-sdk/blob/0cb982c366a90039c00c47639f80632c032ab7b1/common-libs/internal-api-clients/src/services/AffinidiVaultApiService.ts#L30. That is the reason why it is working

sdk/core/test/unit/services/AffinidiVaultStorageService.test.ts
YuriiSkrypnyk: hm, that what I was found in integration test for revocationList `#buildRevocationListStatus, #revokeCredential`
Anyway without it doesn't working. Basically is it using `DidAuthService` from `@affinidi/affinidi-did-auth-lib`, which actually making call to registry.
Do you have some examples how do that without this call?

sdk/core/src/services/WalletStorageService.ts
YuriiSkrypnyk: that is cannot be done with `Set<string>` here as credentials already encrypted and are parsed objects here. I don't really want make stringifying/parsing here to make them unique. Can we distinct them by `id`?

YuriiSkrypnyk: Made this making stringify for each credential and putting it in `map` to check uniqueness.

sdk/core/test/unit/services/AffinidiVaultStorageService.test.ts
Yengas: My bad. It should stay like this. I think in the background this service is called for the flow.

inga-affinidi (dismissed):

inga-affinidi (approved):

Yengas (approved):

PR #118, opened 2021-07-21
FTL-1726

oleksiipiliugin:
FTL-1727. Added AccountOwnership type


sdk/vc-data/src/data/accountOwnership/v1.ts
```diff
@@ -0,0 +1,106 @@
...
+  'AccountOwnershipGithub',
+  {
+    id?: number
+    login?: string
+    url?: string
+    name?: string
+    email?: string
+    followers?: number
+    following?: number
+    twitterUsername?: string
```
inga-affinidi, line 22: Do we need `twitterUsername` in `AccountOwnershipGithub`, or is it a mistype?

oleksiipiliugin: This field from the user's GitHub account might be useful I think. And we can combine it with verification of Twitter account ownership in future

inga-affinidi (approved):

sdk/vc-data/src/data/accountOwnership/v1.test.ts
```diff
@@ -0,0 +1,158 @@
...
+          "https://schema.affinity-project.org/AccountOwnershipV1",
+        ],
+        "https://www.w3.org/2018/credentials#credentialSubject": Array [
+          Object {
+            "https://schema.affinity-project.org/data": Array [
+              Object {
+                "@type": Array [
+                  "https://schema.affinity-project.org/AccountOwnership",
+                ],
+                "https://schema.affinity-project.org/accountName": Array [
```
Yengas, line 53: Some of this fields may use `schema.org` instead of our own type maybe? For example this could be an alias for `https://schema.org/name` or we can just name it as `name`. Same goes for email etc.

Yengas (approved):

DenisPopov15 (approved):

PR #119, opened 2021-07-23
Refactored openapi type mappings in internal-api-clients

inga-affinidi:
No actual code changes in this PR; only types were changed to unblock @oleksiipiliugin


common-libs/internal-api-clients/src/services/AffinidiVaultApiService.ts
```diff
@@ -11,15 +12,15 @@ export type BlobType = {
...
 
 @profile()
-export default class AffinidiVaultApiService extends GenericApiService<ApiSpec> {
+export default class AffinidiVaultApiService extends GenericApiService<ApiType> {
   constructor(options: ConstructorOptions) {
     super(options.vaultUrl, options, affinidiVaultSpec)
   }
 
-  async createDidAuthRequest(params: ExtractRequestType<ApiSpec, 'CreateDidAuthRequest'>) {
+  async createDidAuthRequest(params: ApiType['CreateDidAuthRequest']['requestBody']) {
```
oleksiipiliugin, line 23: IDE ask me to provide return type on function and I think we should do something like that 
 `async createDidAuthRequest(
    params: ApiType['CreateDidAuthRequest']['requestBody'],
  ): Promise<{ body: ApiType['CreateDidAuthRequest']['responseBody']; status: number }> {
    return this.execute('CreateDidAuthRequest', { params })
  }`
@inga-affinidi What do you think?

inga-affinidi: @oleksiipiliugin , it's not IDE, it's our ESLint rules, and I don't think we need this specific rule (which caused to many cases of `: any` because instead of letting TS automatically infer things people had to declare something manually, and `any` is the easiest thing to declare)

vitalyrudenko-affinidi (approved):

common-libs/internal-api-clients/src/services/AffinidiVaultApiService.ts
inga-affinidi: And I had to spend too much time on this issue already (to make parsers more or less universal), so honestly I'm not into doing any polishing right now.

oleksiipiliugin (approved):

PR #120, opened 2021-07-27
FTL-1750 add FileV1 VC type

Yengas:
## Context
https://replika.atlassian.net/browse/FTL-1750

## Description
Added FileV1 VC type for the self-signed VC PoC project. Extended [schema.org/MediaObject](https://schema.org/MediaObject) to be compatible with most of the fields. Only custom field we have added is the `contentAsBase64` which will let us encode and embed binary data into the VC itself for small files.


Volland (dismissed):

sdk/vc-data/package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@affinidi/vc-data",
-  "version": "1.3.2",
+  "version": "1.3.3",
```
inga-affinidi, line 3: You should also update version on line 9

Volland (approved):

sdk/vc-data/package-lock.json
Yengas: ran `npm install` it fixed this and also added some missing dependencies.

Yengas: per inga's comment on Slack(should only use lerna bootstrap in monorepo), I reverted `npm install` changes and manually changed the line 9.

inga-affinidi (approved):

oleksiipiliugin (approved):

DenisPopov15 (approved):

PR #121, opened 2021-07-27
FTL-1757 - added sdk version to request headers

YuriiSkrypnyk:
@affinidi/internal-api-services
- Added `headers` to api services `options`

@affinidi/wallet-core-sdk
- Added helper function `createApiServiceHeaders` to prepare headers with `X-SDK-Version`, which populates from `package.json`
- Passed sdk version header to every api service from `internal-api-services`
- Added request headers to mock in UT
- Fixed integration tests for vault logic to randomize credentials with each tests run. 

image


common-libs/internal-api-clients/src/services/GenericApiService.ts
```diff
@@ -37,7 +37,7 @@ type RequestOptionsForOperation<TApi extends BuiltApiType, TOperationId extends
   RequestOptions<TApi[TOperationId]['requestBody'], TApi[TOperationId]['queryParams'], TApi[TOperationId]['pathParams']>
 >
 
-type ConstructorOptions = { accessApiKey: string }
+export type GenericConstructorOptions = { accessApiKey: string; headers?: NodeJS.Dict<string | string[]> }
```
inga-affinidi, line 40: But it's `additionalHeaders`, not just `headers`.

And the goal of `internal-api-clients` is to isolate everything that has to do with HTTP from its users, so probably better to expect api version here rather than just generic headers.
(And that way we'll also be able to check that api version is always passed)

sdk/core/src/_helpers.ts
```diff
@@ -17,3 +20,9 @@ export const stripParamsFromDidUrl = (did: string): string =>
 export function isW3cCredential(credential: any): boolean {
   return !!credential.type
 }
+
+interface SDKHeaders extends IncomingHttpHeaders {
+  'X-SDK-Version': string
+}
+
+export const createApiServiceHeaders = (): SDKHeaders => ({ 'X-SDK-Version': version })
```
inga-affinidi, line 28: Why do we need to extend `IncomingHttpHeaders` then? SDKHeaders could just contain a single field.

sdk/core/src/_helpers.ts
```diff
@@ -1,6 +1,9 @@
 import { SUPPORTED_DID_METHODS } from './_defaultConfig'
 import SdkErrorFromCode from './shared/SdkErrorFromCode'
 
+import { version } from '../package.json'
```
inga-affinidi, line 4: Do we know that it will work in browser, after everything was minified and bundled etc?

common-libs/internal-api-clients/src/services/GenericApiService.ts
YuriiSkrypnyk: To my opinion header `X-SDK-Version` should come exactly from `core` as version comes exactly from it's `package.json` and only `core` knows about this header.
Since `@affinidi/internal-api-services` is publicly available which means that anyone can use it without `@affinidi/wallet-core-sdk` and shouldn't know anything about sdk version (header), isn't it?

sdk/core/src/_helpers.ts
YuriiSkrypnyk: That made in case we would like to set some predefined headers in future, but can be removed for now and added when we need it.

sdk/core/src/_helpers.ts
YuriiSkrypnyk: Hm, tried that on top of https://github.com/affinityproject/affinidi-demo-browser-app with bumping up sdk version and that puts header correctly, but haven't tried running on minified and bundled version. Will update that here.

YuriiSkrypnyk: that is really didn't work in bundle and seems because of that `bootstrap` stage failed in CodeClimate.
Changed to `require('../package.json')`.

common-libs/internal-api-clients/src/services/GenericApiService.ts
inga-affinidi: We can accept the (required) user-agent string then.
And core can set it to e.g. `AffinidiCoreSdk/6.0.0`.

YuriiSkrypnyk: Not sure understand how user-agent can help here, but the right thing that we should somehow distinguish requests from legacy sdk and without sdk, so then make sense have some default value (for e.g. `X-SDK-Version: unknown`) and move defining this header to `@affinidi/internal-api-services`

sdk/core/test/integration/services/WalletStorageService.test.ts
```diff
@@ -79,7 +84,11 @@ describe('WalletStorageService', () => {
   it.skip('#storeEncryptedSeed throws exception WHEN userId does not exists', async () => {
     const { accessToken } = await userManagementService.logInWithPassword(nonExistingUser, cognitoPassword)
 
-    const keyStorageApiService = new KeyStorageApiService(options)
+    const keyStorageApiService = new KeyStorageApiService({
+      keyStorageUrl: options.keyStorageUrl,
+      accessApiKey: options.accessApiKey,
+      sdkVersion: extractSDKVersion(),
+    })
```
vitalyrudenko-affinidi, line 91: Can't this be moved to `beforeEach`? They look exactly the same and now it's more visible that it's just a repeated code.

common-libs/internal-api-clients/src/helpers/headers.ts
```diff
@@ -0,0 +1,30 @@
...
+    'X-SDK-Version': options.sdkVersion || defaultHeaderValue,
+  }
+}
+
+type updateHeadersOptions = {
+  authorization?: string
+  storageRegion?: string
+}
+
+export const updateHeaders = (headers: ApiRequestHeaders, options: updateHeadersOptions) => {
```
inga-affinidi, line 24: The name is misleading, as if this method changes the input value.

An example of a better name would be `getExtendedHeaders`.

common-libs/internal-api-clients/test/unit/helpers/headers.test.ts
```diff
@@ -0,0 +1,69 @@
...
+    expect(headers['X-SDK-Version']).to.eql('unknown')
+  })
+
+  it('#updateHeaders', async () => {
+    let headers = createHeaders({
+      accessApiKey: fakeAccessApiKey,
+      sdkVersion: fakeSDKVersion,
+    })
+
+    headers = updateHeaders(headers, {
```
inga-affinidi, line 41: But we don't need to change the original variable here?
It could be `const originalHeaders` and `const extendedHeaders` as well.

sdk/core/test/integration/services/WalletStorageService.test.ts
YuriiSkrypnyk: That does not needed for every test in suite to put it into `beforeEach` and don't want that each test could use same instance of `KeyStorageApiService` service created in `beforeAll`. But we actually can move service creation into separate function which removes repeated lines.

inga-affinidi (approved):

common-libs/internal-api-clients/src/services/GenericApiService.ts
inga-affinidi: User-agent can help in such a way that both SDK and other `internal-api-users` can pass something reasonable there, while `X-SDK-Version`, as you have mentioned, is specific to SDK.

vitalyrudenko-affinidi (approved):

PR #122, opened 2021-07-29
FTL-1691 User management refactoring

inga-affinidi:
## Checklist
- [x] `dangerous` exports removed
- [x] Cloud Wallet Api: all unit, integration and functional tests pass (after code was upgraded to use new SDK; branch `il-sdk-v6-beta-compatibility`; besides some bugs related to affinidi vault which are outside of this scope, cloud-wallet-api was never tested with affinidi vault)
- [x] Versions bumped
- [x] Changelog updated
- [x] Readme updated
- [x] Pull request description written

## Low-effort list of changes:
* Duplicate methods from platform SDKs moved out to core-sdk 
* Tests, CI
  * Duplicate tests from platform SDKs moved to core 
  * OTP tests are now run in parallel
  * Tests (and other build scripts) are run in parallel across packages
  * Streamlined build script definitions (see readme.md), now they have the same meaning across packages
  * Removed redundant steps from github actions (tests were run twice)
  * Total github build time is 10-12 minutes, down from 30 minutes before, despite 2x increase in number of OTP tests (one set of OTP tests checks the legacy code, and another, v6 code)
* SdkOptions reexported directly from platform SDKs (was in dist/AffinityWallet)
* WalletFactories instead of CommonNetworkMember
  * AffinidiWallet is no longer a class
  * (although you still can invoke it with `new`, for backward compatibility)
  * Classes are no longer exported, you cannot extend or override
* Legacy and v6
  * Legacy static methods now return legacy class instances which have legacy methods
  * New static methods return new class instances without legacy methods (e.g. new instance has `logOut()`, legacy instance has `signOut(options)`)
  * Full backwards compatibility (with caveats); however, legacy methods will be removed in v7
  * Also new instances are context-dependent: if you created an instance with `createFromEncryptedSeedAndPassword`, it won't have `logOut` method
  * Do not want legacy methods to pollute your IDE? `AffinidiWalletV6` only contains non-legacy methods
  * Major refactoring of the user management methods, now they have uniform names and logic:
    * Methods that always have immediate effect (`signUpWithUsername`, `logInWithPassword`, `changePassword`, `logOut`)
    * Methods that always require confirmation (`signUpByEmail/Phone`, `logInPasswordless`, `signInPasswordless`, `forgotPassword`, `changeEmail/Phone`) come in pairs: `initiateAction` accepts parameters and returns token, `completeAction` accepts token and confirmation code. Only `initiate...` methods accept `MessageParameters`
    * For convenience, `completeForgotPassword` now logs user in and returns a wallet nistance
    * Have some keys already? Just pass them to `signUp` or `confirmSignUp` as optional parameters; no more duplicate methods
    * Factories always accept `SdkOptions` as a first argument (was on different positions previously), instance methods never accept `SdkOptions` as an argument (some required it before)
    * Need to get a string token so that you could reinstate wallet instance later? Instead of manually handling cognito access tokens, use instance method `serializeSession` and static method `deserializeSession`
    * New methods are covered by tests derived from ones for legacy methods; legacy methods are covered by original tests
    * There is also a new method `Util.getLoginType`, it returns `'email'` or `'phone'` or `'username'`; if you relied `AffinidiWallet.signUp` in older versions of SDK returning `AffinidiWallet | string` and detecting resulting value type to decide whether to ask user for a confirmation code, you now can do that prior to calling signup (and decide whether to call `signUpWithUsername` or `signUpByEmail`)
  * Static methods that do not depend on AffinidiWallet now exist on `Util` (for backwards compatibility they're still available on `AffinidiWallet` until v7)
  * No more `CommonNetworkMember` with 2k LOC; it is split into separate files, and the largest file remaining is `BaseNetworkMember` at 1152 LOC (to be split into smaller classes in the future)
* New exports, to make `wallet-{PLATFORM}-sdk` your single point of entry:
  * New methods in `AffinidiWallet`: `signUnsignedCredential`, `signUnsignedPresentation`, `signJwt`
  * And in `Util`: `getDidFromToken`, `getPublicKeyHexFromDidDocument`, `generateSeed`, `fromJWT`, `getLoginType`
  * Everything `dangerous` is removed
  * `FreeFormObject` is removed (it produced more problems than it solved), instead `any` is used
  * A bunch of types are exported from the main entry point of SDK
  * SDK clients will only have to `import ... from '@affinidi/wallet-{PLATFORM}-sdk` now. No longer importing from `@affinidi/wallet-core-sdk` or `@affinidi/common` or `@affinidi/vc-common` or `@affinidi/wallet-browser-sdk/dist/dto/shared.dto` etc.
* Implemented limited BBS support as a part of https://replika.atlassian.net/browse/FTL-1707
  * The code is very low-quality full of hacks and workarounds; it will be improved as a part of https://replika.atlassian.net/browse/FTL-1707 when `affinidi-vc-signatures` code is migrated to SDK
* Removed unneeded dependencies


sdk/core/src/CommonNetworkMember/NetworkMemberWithoutCognito.ts
```diff
@@ -0,0 +1,98 @@
...
+}[keyof T]
+type OmitConstructor<T> = Omit<T, ConstructorKeys<T>>
+type DerivedTypeForOptions<TInstance> = Constructor<TInstance> &
+  AbstractStaticMethods &
+  OmitConstructor<typeof NetworkMemberWithoutCognito>
+type DerivedType<T extends DerivedType<T>> = DerivedTypeForOptions<InstanceType<T>>
+export type UniversalDerivedType = DerivedType<DerivedTypeForOptions<NetworkMemberWithoutCognito>>
+
+@profile()
+export abstract class NetworkMemberWithoutCognito extends BaseNetworkMember {
```
vitalyrudenko-affinidi, line 24: Why do we need to add a suffix if class doesn't contain something? Why not "NetworkMember" and "NetworkMemberWithCognito"? It'll be too hard to use?

sdk/core/src/CommonNetworkMember/Util.ts
```diff
@@ -0,0 +1,115 @@
...
+   * @description Parses JWT token (request and response tokens of share and offer flows)
+   * @param token - JWT
+   * @returns parsed object from JWT
+   */
+  fromJWT: (token: string): any => {
+    // await ParametersValidator.validateSync(
+    //   [
+    //     { isArray: false, type: 'jwt', isRequired: true, value: token }
+    //   ]
+    // )
```
vitalyrudenko-affinidi, line 73: Why are these commented out?

sdk/core/src/factories/legacyWalletFactories.ts
```diff
@@ -0,0 +1,317 @@
...
+      cognitoUserTokens?: CognitoUserTokens,
+    ) {
+      super(password, encryptedSeed, platformEncryptionTools, affinidiCommon, options, component, cognitoUserTokens)
+    }
+  }
+
+  return Wallet as UniversalDerivedType
+}
+
+export const createLegacyWalletFactories = (
```
vitalyrudenko-affinidi, line 31: Where are the factories? What are these functions do, exactly? They look a bit confusing to me without context

sdk/core/src/shared/interfaces.ts
```diff
@@ -1,10 +1,76 @@
...
+ */
+type AffinidiCommonConstructorOptions = {
+  apiKey: string
+  registryUrl: string
+  metricsUrl: string
+  component: EventComponent
+}
+
+/**
+ * @deprecated Temporary implementation; refactor by 6.0 release (FTL-1707)
```
vitalyrudenko-affinidi, line 74: How it would look after being refactored?

sdk/browser/src/AffinityWallet.ts
```diff
@@ -1,82 +1,15 @@
...
-    return this.getAllCredentials()
-  }
-}
+export const AffinidiWallet = createV5CompatibleWalletFactories(
+  platformEncryptionTools,
+  null,
+  EventComponent.AffinidiBrowserSDK,
+)
+
+export const AffinidiWalletV6 = createV6WalletFactories(
```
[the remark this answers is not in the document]
inga-affinidi: Because of the backwards compatibility

README.md
```diff
@@ -66,3 +66,16 @@ Before we begin, make sure you have the following installed:
...
+The following commands should be supported by all projects in monorepo
+
+* `prepare`: should build package
+* `lint`: should run lint
+* `checkLicenses`: should check licenses
+* `test:unit`: should run unit tests, if any
+* `test:integration`: should run integration tests, if any
+* `test`: should run all tests
+* `test:coverage`: should run all tests with coverage
+* `test:coverage:codeclimate`: should run `test:coverage` and save codeclimate results
```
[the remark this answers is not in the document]
inga-affinidi: I intended these to be requirements, not descriptions :)

sdk/core/src/CommonNetworkMember/NetworkMemberWithoutCognito.ts
inga-affinidi: If we had `NetworkMemberWithCognito` and `NetworkMember` it would look like `NetworkMemberWithCognito` extends `NetworkMember`?

These are two different classes extended from the same base; one handles scenarios when user management is required, and another handles scenarios where there should be no user management at all.

Maybe the current naming is questionable; do you have any ideas besides just plain `NetworkMember`?

sdk/core/src/CommonNetworkMember/Util.ts
inga-affinidi: This is a verbatim copy of the existing legacy method, I didn't have time to refactor everything :)

sdk/core/src/shared/interfaces.ts
inga-affinidi: It will be deleted after refactoring. The only reason for it to exist is that we need to somehow inject `AffinidiSignatures` into several NetworkMember methods, because `AffinidiSignatures` is defined in a private package and cannot be used in SDK.

sdk/core/src/factories/legacyWalletFactories.ts
inga-affinidi: Originally, SDK exported a class with a bunch of static methods returning instances of that class:
```
class CommonNetworkMember {
  constructor(platformEncryptionTools, ...otherParams) {}

  static fromLoginAndPassword(login, password) { return new this(...otherParams) }
}

// in react-native
export class AffinidiWalletReactNative extends CommonNetworkMember {
  constructor(...otherParams) {
    super(reactNativePlatformTools, ...otherParams)
  }
}

// usage
const wallet = AffinidiWalletReactNative.fromLoginAndPassword(...) // here, AffinidiWalletReactNative is a class, and wallet is an instance of that class
```

Now instead of exporting a class, we have
```
const createLegacyWalletFactories = (platformEncryptionTools) => ({
  fromLoginAndPassword: (login, password) => //returns CommonNetworkMember
})

const createV6Factories = (platformEncryptionTools) => ({
  logInWithPassword: (login, password) => //returns CommonNetworkMember
})

const createFactories = // combines legacy and v6 factories

// in react-native
export const AffinidiWalletReactNative = createFactories(reactNativePlatformTools)

// usage
const legacyWallet = AffinidiWalletReactNative.fromLoginAndPassword(...) // here, AffinidiWalletReactNative is just an object, fromLoginAndPassword is its field, and legacyWallet is an instance of internal "legacy wallet" class defined in core sdk only

const walletV6 = AffinidiWalletReactNative.logInWithPassword(...) // walletV6 is an instance of internal "wallet with user management" class defined in core SDK only
```

This allowed me to split CommonNetworkMember.ts into several smaller files, and to have three different leaf classes (legacy, with cognito, without cognito) each with its own set of methods and with its own set of constructor parameters.
So that you can do `legacyWallet.signOut(options)` or `walletV6.logOut()` but not vice versa; and you cannot call `signOut` or `logOut` on an object returned by `AffinidiWalletReactNative.createFromEncryptedSeedAndPassword`.

.github/workflows/public-packages-publish-workflow.yaml
```diff
@@ -36,13 +36,6 @@ jobs:
           TEST_SECRETS: ${{secrets.INTEGRATION_TEST_SECRETS}}
           TESTMAIL_API_KEY: ${{secrets.TESTMAIL_API_KEY}}
           TESTMAIL_NS: ${{secrets.TESTMAIL_NS}}
-      - name: Run integration tests
```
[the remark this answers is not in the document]
inga-affinidi: This flow already invokes `npm test`, and according to the new readme (in the root directory), `npm test` should execute all tests, both unit and integration ones.

While previously it was a mess, `npm test` executed all tests for some packages, lint + all tests for some others, lint + unit tests for yet other, etc.

In this PR, all commands have clearly defined meaning across all packages: https://github.com/affinityproject/affinidi-core-sdk/pull/122/commits/fb7d4ee6cd973f8d26e131c2b8501e59712d6bbc

sdk/core/src/CommonNetworkMember/BaseNetworkMember.ts
```diff
@@ -0,0 +1,1151 @@
...
+   * 5. anchor DID with DID document ID from IPFS
+   * @param password - encryption key which will be used to encrypt randomly created seed/keys pair
+   * @param options - optional parameter { registryUrl: 'https://affinity-registry.dev.affinity-project.org' }
+   * @returns
+   *
+   * did - hash from public key (your decentralized ID)
+   *
+   * encryptedSeed - seed is encrypted by provided password. Seed - it's a source to derive your keys
+   */
+  protected static async _register(
```
Volland, line 177: @inga-affinidi @DenisPopov15 should we make register as a public method? We had a chat about it few days ago

Yengas: Isn't the point of these changes move user/login/auth related stuff out of the base network member and related core classes? Maybe this function can be taken out to another class instead of making it available publicly in the BaseNetworkMember?

Yengas (dismissed):

sdk/core/src/dto/shared.dto.ts
```diff
@@ -164,11 +173,11 @@ export class CredentialParams {
 
   @IsDefined()
   @IsArray()
-  context: FreeFormObject[]
+  context: unknown[]
```
Yengas, line 176: what problems we were having with freeformobject? and why did we remove it? removing it can cause backwards incompatible changes where we return these types. and the client will need to do type checks / guards where he was previously accessing FreeFormObjects wthout any checks.

inga-affinidi: I've removed `FreeFormObject` because it was not worth it. It increases the validation complexity significantly, is recursive (of arbitrary depth), often does not cover some cases (e.g. some field that is usually a `string[]` was declared as `FreeFormObject` but we didn't have type checking at the moment; after I enabled type checking for that field (by removing some `as any`), it started to fail).

It is exported and then used in services (such as cloud-wallet-api) for tsoa validation, which means that it is not even typescript, that it should be written in tsoa-compliant subset of typescript (so SDK would have to know implicitly about tsoa; and in the future a developer could introduce a benign change to the SDK which would break tsoa compatibility without that developer even suspecting).

In most of the cases, data typed as `FreeFormObject` is subject to some more specific validation later according to its specific schema anyway, so not much sense in validating this data to be `FreeFormObject` beforehand.

And after trying to cover all cases, which would look like this:
```type FreeFormObject = string | FreeFormObject[] | Record<string, FreeFormObject>```
which basically means that it is something compatible with JSON and not containing any bools or numbers.
Are we sure that data typed as FreeFormObject should not contain bools or numbers?
Is it worth it to maintain all that complexity just to have this restriction declared on TS level (even though the specific data objects will still be validated by their specific schemas)?

Comment (inga-affinidi) [reply sdk/core/src/dto/shared.dto.ts:176]: > removing it can cause backwards incompatible changes where we return these types.

This is an input parameter, and clients can supply any array (including an array of former `FreeFormObject`s) to a method expecting `unknown[]`.

**Review by Yengas — APPROVED** [review]

File: sdk/core/src/CommonNetworkMember/BaseNetworkMember.ts (hunk 8 above)
Comment (inga-affinidi) [reply sdk/core/src/CommonNetworkMember/BaseNetworkMember.ts:177]: @Volland @Yengas , the public method is `register` in `LegacyNetworkMember` and `createWallet` in `NetworkMemberWithoutCognito`

File: sdk/core/CHANGELOG.md [hunk 10]
```diff
@@ -1,3 +1,77 @@
...
+* Introduced new streamlined set of factories and methods for user management, with clearer return values.
+  - Note that instances returned by non-legacy factories do not have legacy methods, and vice versa.
+  - Note that in new methods, every method accepting confirmation code now also requires to be passed a token,
+    which is the string value returned by a method that sent a confirmation code.
+  - Note that new instance (non-static) methods no longer accept an options object.
+  - Note that signup methods that return "either `string` or `AffinidiWallet`" are deprecated,
+    instead you should either let user to explicitly choose between using an username, email address or phone number,
+    or detect the type of login by invoking `Util.getLoginType`,
+    and then invoke the relevant signup method (`signUpWithUsername` or `initiateSignUpByEmail` or `initiateSignUpByPhone`).
+* Implemented new methods:
```
DenisPopov15, line 13: @inga-affinidi I believe it's not a full list of new methods, correct? So I am as a consumer, I can get that all my methods deprecated, but what I should use instead? I believe at least we should add a reference to the README (e.g. - new interface check here)

DenisPopov15 (approved):

Yengas (approved):

Volland (approved):

sdk/core/CHANGELOG.md
inga-affinidi: @DenisPopov15 , deprecated methods are marked as deprecated with `@deprecated` comments pointing to the specific non-deprecated alternatives.

PR #123, opened 2021-07-30
Added did auth abstract service

YuriiSkrypnyk:
- Created `DidAuthApiService` abstract class, which is middleware between API services based on Did authentication.
- Moved out DID auth related code from `AffinidiVaultStorageService`
- Upgraded SDK version
- Added check for the token expiration
- Caching token for particular audience DID to be reused


common-libs/internal-api-clients/src/services/DidAuthApiService.ts
```diff
@@ -0,0 +1,60 @@
...
+    const didAuthRequestToken = await this.execute(this._getDidAuthOperationId, {
+      params: {
+        audienceDid: this._audienceDid,
+      },
+    } as RequestOptionsForOperation<TApi, TDidAuthOperationId>)
+    return await this._didAuthService.createDidAuthResponseToken(didAuthRequestToken.body)
+  }
+
+  private async _getToken() {
+    if (!this._responseToken || !this._didAuthService.isTokenExpired(this._responseToken)) {
```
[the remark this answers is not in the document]
Yengas: We should have a test for this.

Yengas (changes requested):

common-libs/internal-api-clients/package.json
```diff
@@ -49,6 +49,7 @@
   "devDependencies": {
     "@affinidi/eslint-config": "1.0.1",
     "@affinidi/prettier-config": "1.0.1",
+    "@affinidi/affinidi-did-auth-lib": "^1.0.7",
```
inga-affinidi, line 52: But it is used as a non-dev dependency

common-libs/internal-api-clients/src/services/DidAuthApiService.ts
```diff
@@ -0,0 +1,60 @@
...
+import GenericApiService, { GenericConstructorOptions } from './GenericApiService'
+
+export type DidAuthConstructorOptions = GenericConstructorOptions & {
+  audienceDid: string
+  didAuthService: DidAuthService
+}
+@profile()
+export default abstract class DidAuthApiService<
+  TApi extends BuiltApiType,
+  TDidAuthOperationId extends keyof TApi
```
inga-affinidi, line 15: We should only allow operations that are compatible with what we expect did auth operation to look like.

So it should be not `keyof TApi` but rather only keys such that `TApi[key]['response']` extends the expected response, expected request extends `TApi[key]['request']`, etc.

inga-affinidi:
As a part of this PR we should probably also get rid of `@affinidi/affinidi-did-auth-lib` explicit dependency in core-sdk.

common-libs/internal-api-clients/src/services/DidAuthApiService.ts
Yengas: You mean: we need to filter out only the endpoints that have did-auth from the `TApi` definition? E.g. only endpoints that require `bearerAuth` security?

inga-affinidi: I mean, we need to filter out only the endpoints that accept `{ audienceString: string }` and return `string`. So that you'll have typechecking in `_createDidAuthToken`

Yengas: I've changed this to only accept operation ids that look like create did auth requests. However, the inner method still has type casting. I don't mind this and since the public API is better than before, and writing more TS would only change internals of this class, I would prefer not to spend more time on this and don't want to complicate the types.

But If you think we should still improve the types, I would be happy to pair and try to improve it. LMK what you think.

common-libs/internal-api-clients/package-lock.json
```diff
@@ -38,6 +39,59 @@
         "typescript": "^4.3.2"
       }
     },
+    "node_modules/@affinidi/affinidi-did-auth-lib": {
+      "version": "1.0.8",
+      "resolved": "https://registry.npmjs.org/@affinidi/affinidi-did-auth-lib/-/affinidi-did-auth-lib-1.0.8.tgz",
+      "integrity": "sha512-v0RE14H+CLWZ4PwVNCX+xgF5Koqw14wdt8n1KoFQfiI440wduqZXMinDuwIopY40hZ0XxKTDaGBxvlTnntjJgQ==",
+      "dependencies": {
+        "@affinidi/common": "^1.3.1",
```
inga-affinidi, line 47: We _really_ need to fix this someday...

(Probably by making `affinidi-did-auth-lib` a part of this monorepo)

inga-affinidi (approved):

Yengas (approved):

DenisPopov15 (approved):

PR #125, opened 2021-08-10
feat: Add Hackathon VC with awardDate

hiteshjoshi1:
- Hackathon Winner VC
- uses schema.orh/Hackathon
- Just adds one new field awardDate

### DIFF
--- a/sdk/vc-data/src/data/hackathon/v1.test.ts
+++ b/sdk/vc-data/src/data/hackathon/v1.test.ts
@@ -1,0 +1,142 @@
+import { VCHackathonWinnerV1, VCSHackathonWinnerV1, getVCHackathonWinnerV1Context } from './v1'
+import { expandVC } from '../../testUtil.test'
+
+describe('VCGenderPersonV1', () => {
+  it('expands correctly', async () => {
+    expect.assertions(1)
+
+    const expanded = await expandVC<VCHackathonWinnerV1, VCSHackathonWinnerV1>({
+      type: 'HackathonWinnerCredentialV1',
+      data: {
+        '@type': ['Hackathon', 'HackathonE', 'HackathonWinner'],
+        name: 'Insanely Glorified Hackathon',
+        organizer: {
+          legalName: 'Midas Touch Pte Ltd',
+          employee: {
+            name: 'Slatirbartfast',
+            email: 'slatir@midas.sg',
+            jobTitle: 'Managing Director',
+          },
+        },
+        funder: [
+          {
+            name: 'Crypto Funds.com',
+          },
+        ],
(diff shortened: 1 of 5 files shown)


sdk/vc-data/src/data/hackathon/v1.test.ts
```diff
@@ -0,0 +1,142 @@
...
+                      },
+                    ],
+                    "https://schema.org/leiCode": Array [
+                      Object {
+                        "@value": "A16Z112s",
+                      },
+                    ],
+                  },
+                ],
+                "https://schema.org/awardDate": Array [
```
Yengas, line 69: there is no `awardDate` on schema.org. Would prefer to use `affSchema` or use some other field for storing the same information. E.g. `hackathon.workPerformed` has `dateCreated, dateModified, datePublished` fields.

hiteshjoshi1: I initially also went with hackathon.datePublished  but then changed my mind because awardDate is the date on which the fintech is awarded the Hackathon winner credential and is not technically a part of workPerformed.
That said, I do not have strong feelings about using awardDate vs say datePublished. 
However, if I do use awardDate, what will I have to change in v1.ts if i want to get awardDate from affSchema.  I am newbie and doing this for first time. 
I tried specifying 
```
awardDate:affSchema
```
But that generates in the test
```
“https://schema.org/undefinedawardDate”: Array [
                  Object {
                    “@value”: “26-14-2021”,
                  },
                ]```

Yengas: From the test output you can see that you are actually writing types for the `credentialSubject.data` part of the credential. Please check [this example VC](https://www.w3.org/TR/vc-data-model/#example-1-a-simple-example-of-a-verifiable-credential) to see other standard fields besides the **credentialSubject**. After you create your type and sign a VC, that VC will have similar fields.

1. For that reason if **awardDate** means the date that the Award was issued by the issuer for the holder, then the **issuanceDate** can serve the same purpose. This means it maybe unnecessary to add it.
2. If **awardDate** means something else, we can put it under the `https://schema.affinity-project.org/awardDate` using [this syntax](https://github.com/affinityproject/affinidi-core-sdk/blob/master/sdk/vc-data/src/data/media/v1.ts#L45).

hiteshjoshi1: I updated the PR , keeping the award date for now but changed it to https://schema.affinity-project.org/awardDate

Yengas (approved):

oleksiipiliugin (approved):

PR #129, opened 2021-08-18
Allow did:elem to be built from external key vaults

cosminm-affnd:
Extract key management from ElemDidDocument


DenisPopov15 (approved):

common-libs/common/src/services/DidDocumentService/KeyVault.ts
```diff
@@ -0,0 +1,26 @@
+/**
+ * Abstraction of a key vault used for generating did documents
+ */
+export interface KeyVault {
+  /**
+   * Primary public key
+   */
+  primaryPublicKey(): Buffer
```
inga-affinidi, line 8: This should either be a getter property, or if we want it to be a function, the name should look like `getPrimaryPublicKey`

common-libs/common/src/services/DidDocumentService/LocalKeyVault.ts
```diff
@@ -0,0 +1,46 @@
+import { KeyVault } from './KeyVault'
+import { KeysService } from '../index'
+import * as secp256k1 from 'secp256k1'
+
+/**
+ * KeyStore backed implementation when the seed is known
+ */
+export class LocalKeyVault implements KeyVault {
```
inga-affinidi, line 8: Will there be any other keyvaults?

inga-affinidi (changes requested):
Please also take a look at `KeyManagementService` in `sdk\core`, it could make sense to refactor it to use `KeyVault` instead of `KeysService` as well

common-libs/common/src/services/DidDocumentService/LocalKeyVault.ts
cosminm-affnd: Yes that's the plan, e.g in the web3 experiment we'll use metamask as the provider of keys.

common-libs/common/src/services/DidDocumentService/LocalKeyVault.ts
```diff
@@ -0,0 +1,46 @@
...
+  private _keyService: KeysService
+
+  constructor(keyService: KeysService) {
+    this._keyService = keyService
+  }
+
+  primaryPublicKey(): Buffer {
+    const { seed } = this._keyService.decryptSeed()
+    const seedHex = seed.toString('hex')
```
inga-affinidi, line 17: Does it really make sense to call `decryptSeed` on every `KeyVault` method call?

For example, prior to that refactoring, `ElemDidDocument._getMyDidConfig` only called `decryptSeed` once; now it is four indirect calls (via keyvault)

cosminm-affnd:
> Please also take a look at `KeyManagementService` in `sdk\core`, it could make sense to refactor it to use `KeyVault` instead of `KeysService` as well

@inga-affinidi  I agree, ideally we only have one interface around key storage. Including that change here will explode the scope of this MR, to keep things manageable we'll create a MR per service that needs to be updated. I'll add a ticket in our backlog for it.

common-libs/common/src/services/DidDocumentService/elem-lib/func.ts
```diff
@@ -38,12 +38,13 @@ export const getDidUniqueSuffix = (operation: any) => {
 }
 
 // NOTE check is signatures are the same as sidetree's (see NEP-334)
-export const signEncodedPayload = (encodedHeader: string, encodedPayload: string, privateKey: string) => {
+export const signEncodedPayload = (
+  encodedHeader: string,
+  encodedPayload: string,
+  signFn: (payload: Buffer) => Buffer,
+): Buffer => {
```
inga-affinidi, line 45: Seems like it could be `getEncodedPayloadHash` instead, with one less level of indirection.

So that it would be called as
```
const hash = getEncodedPayloadHash(...)
const signed = sign(hash, ...)
```
instead of
```
const signed = signEncodedPayload(..., (hash) => sign(hash, ...))
```

common-libs/common/src/services/DidDocumentService/elem-lib/op.ts
```diff
@@ -18,20 +19,20 @@ export const getDidDocumentModel = (primaryPublicKey: string, recoveryPublicKey:
...
     operation: 'create',
     kid: '#primary',
     alg: 'ES256K',
   }
-  return makeSignedOperation(header, didDocumentModel, primaryKey.privateKey)
+  return makeSignedOperation(header, didDocumentModel, signFn)
 }
 
-export const makeSignedOperation = (header: any, payload: any, privateKey: any) => {
+export const makeSignedOperation = (header: any, payload: any, signFn: (payload: Buffer) => Buffer) => {
```
inga-affinidi, line 32: Similarly, this method could return `{ encodedHeader, encodedPayload }`, and let its callers to call `signFn`.

common-libs/common/src/services/DidDocumentService/ElemDidDocument.ts
```diff
@@ -78,36 +77,15 @@ export default class ElemDidDocument {
     }
   }
 
-  private _buildDIDDocModel(
-    seedHex: string,
-    publicKeyPrimary: string = null,
-    privateKeyPrimary: string = null,
-    publicKeyRecovery: string = null,
-    privateKeyRecovery: string = null,
```
inga-affinidi, line 86: So apparently the old version of this method didn't actually use private keys. @DenisPopov15 , could it be that there was some bug and private keys should be somehow used here after all?

common-libs/common/src/services/DidDocumentService/LocalKeyVault.ts
cosminm-affnd: That's a good question, what was the original thinking around keeping the key encrypted in memory? It does not seem to improve security by much, as both the encrypted key and secret are stored next to each other.

The easiest solution is to keep the seed in clear text in memory to avoid repeated decryptions. A slightly more complex way would be to add an `unlock()/lock()` mechanism to `KeyVault`. 

@DenisPopov15 you also raised this, any preferrences?

common-libs/common/src/services/DidDocumentService/LocalKeyVault.ts
```diff
@@ -0,0 +1,46 @@
...
+  externalKeys(): any[] {
+    const { externalKeys } = this._keyService.decryptSeed()
+    return externalKeys
+  }
+
+  sign(payload: Buffer): Buffer {
+    const { seed } = this._keyService.decryptSeed()
+    const seedHex = seed.toString('hex')
+
+    const { privateKey } = KeysService.getPublicAndPrivateKeys(seedHex, 'elem')
```
inga-affinidi, line 40: We might want to cache the result of `getPublicAndPrivateKey` in this class. This operation seems to be somewhat time-consuming; and these changes introduce performance degradation, as we only called `getPublicAndPrivateKeys` once in the original version.

common-libs/common/src/services/DidDocumentService/LocalKeyVault.ts
inga-affinidi: Also what is the point of making it a class instead of a plain object, if you have declared `KeyVault` interface anyway?

```
export const createKeyVault = (keysService: KeysService) => ({
  getPrimaryPublicKey: () => ...,
  ...
})
}

File: common-libs/common/src/services/DidDocumentService/elem-lib/op.ts [hunk 8]
```diff
@@ -18,20 +19,20 @@ export const getDidDocumentModel = (primaryPublicKey: string, recoveryPublicKey:
...
   const encodedHeader = encodeJson(header)
   const encodedPayload = encodeJson(payload)
-  const signature = signEncodedPayload(encodedHeader, encodedPayload, privateKey)
+  const signature = base64url.encode(signEncodedPayload(encodedHeader, encodedPayload, signFn))
```
inga-affinidi, line 35: The original version of the code called `base64url.encode` on `signatureObject.signature`. The new one calls it on `Buffer.from(signatureObject.signature)`. Not sure what is the point of this change, and not sure if it doesn't break backwards compatibility?

common-libs/common/src/services/DidDocumentService/LocalKeyVault.ts
inga-affinidi: > what was the original thinking around keeping the key encrypted in memory? It does not seem to improve security by much

What's more, once it is decrypted, it is stored in memory in cleartext anyway until (if) garbage collector collects it.

So with these changes, the number of cleartext seed copies in memory is potentially increased 4x.

common-libs/common/src/services/DidDocumentService/elem-lib/op.ts
cosminm-affnd: The type of parameters changed from `string` to `Buffer` to avoid some unnecessary encoding/decoding of hex string to buffer.

The tests are still passing, so backward compatibility is still there. Are there other tests that are missing that need to be added?

inga-affinidi: But in this case it added unnecessary decoding. The original version executed `base64url.encode` directly on `signatureObject.signature`, while the new one additionally calls `Buffer.from` (in keyvault) prior to `base64url.encode`.

> The tests are still passing, so backward compatibility is still there.

Generally tests only check that the current version correctly decodes what it encoded etc.

Do we have some special tests in this case that would check that the old version correctly decodes what the new version has encoded and vice versa?

common-libs/common/src/services/DidDocumentService/ElemDidDocument.ts
```diff
@@ -1,17 +1,16 @@
 import base64url from 'base64url'
 import { parse } from 'did-resolver'
 
-import { op, func } from './elem-lib'
-
-import KeysService from '../KeysService'
+import { func, op } from './elem-lib'
+import { KeyVault } from './KeyVault'
 
 export default class ElemDidDocument {
```
inga-affinidi, line 7: Do I understand it right that this PR is only supposed to refactor things and make code more maintainable, without changing the existing or adding some new behavior?

common-libs/common/src/services/DidDocumentService/LocalKeyVault.ts
cosminm-affnd: Ease of use, It's a common use case to use the local seed. In this case all a client needs to do is call `new LocalKeyVault`

common-libs/common/src/services/DidDocumentService/KeyVault.ts
cosminm-affnd: Do we have code style conventions around this that I could follow?

common-libs/common/src/services/DidDocumentService/LocalKeyVault.ts
cosminm-affnd: Are you suggesting we should use the clear text version?

common-libs/common/src/services/DidDocumentService/ElemDidDocument.ts
cosminm-affnd: Yes, small change that enables the metamask experiment and makes the code easier to maintain.

common-libs/common/src/services/DidDocumentService/LocalKeyVault.ts
inga-affinidi: That seems like most reasonable solution here. Or maybe there are some other drawbacks?

common-libs/common/src/services/DidDocumentService/LocalKeyVault.ts
inga-affinidi: They could call `createKeyVault` all the same

common-libs/common/src/services/DidDocumentService/KeyVault.ts
inga-affinidi: I'm not sure we have some formal conventions at affinidi, but that's a widely-accepted practice and the existing code at affinidi largely follows it.

For example, google code style explicitly mentions it: https://google.github.io/styleguide/jsguide.html#naming-method-names
or here is a mention of that practice in airbnb: https://github.com/airbnb/javascript/issues/848

common-libs/common/src/services/DidDocumentService/elem-lib/op.ts
cosminm-affnd: https://github.com/affinityproject/affinidi-core-sdk/blob/master/common-libs/common/test/unit/services/DidDocumentService.test.ts#L121 Explicitly compares the output with a previously generated did, is it not sufficient?

I'll change the `sign` method to return `Uint8Array` which makes more sense than Buffer, we don't expect to edit the signature.

common-libs/common/src/services/DidDocumentService/KeyVault.ts
cosminm-affnd: Unrelated, but please add this to the SSDLC document.

common-libs/common/src/services/DidDocumentService/LocalKeyVault.ts
cosminm-affnd: Back to coding style, I'm happy to use any approach the team agrees on.

common-libs/common/src/services/DidDocumentService/LocalKeyVault.ts
cosminm-affnd: I'm closing this as it's similar to https://github.com/affinityproject/affinidi-core-sdk/pull/129#discussion_r692002993

common-libs/common/src/services/DidDocumentService/elem-lib/op.ts
cosminm-affnd: The change to `Uint8Array` still requires `Buffer.from` in other parts of the code to transform the buffer to hex, let's leave it as is for now.

common-libs/common/src/services/DidDocumentService/ElemDidDocument.ts
DenisPopov15: @inga-affinidi yeap, Probably its just over required param here (so not really used), the only place where potential its could be required here - its to sign didDocument

DenisPopov15:
> Please also take a look at `KeyManagementService` in `sdk\core`, it could make sense to refactor it to use `KeyVault` instead of `KeysService` as well

@inga-affinidi Are you talking about KeyVault which incapsulate work with key vault storage ?
If so, I would vote do not change this staff, cause this service its more low level thing, which could work with any provider, and that keyVault which you mentioned could one of provider

common-libs/common/src/services/DidDocumentService/LocalKeyVault.ts
DenisPopov15: yeap, since anyway 2 parts of secret (encrypteddata and encryption key in memory anyway) - I believe we could cache it there then in open way.
P.S. I believe we should create a ticket to make research and try resolve this situation on our backlog (do not keep seed in open way or encryptedSeed with encryptionKey together in memory), but this is separate topic and will require some time to research

common-libs/common/src/services/DidDocumentService/elem-lib/func.ts
cosminm-affnd: Moved the functions in op to DidElemDocument where they have access to the vault

common-libs/common/src/services/DidDocumentService/LocalKeyVault.ts
cosminm-affnd: For now this is working let's update it when needed.

common-libs/common/src/services/DidDocumentService/KeyVault.ts
```diff
@@ -0,0 +1,26 @@
...
+
+  /**
+   * Recovery public key if available
+   */
+  recoveryPublicKey: Buffer
+
+  /**
+   * Array of external keys to be added to the elem object
+   */
+  externalKeys(): any[]
```
inga-affinidi, line 18: You forgot to make this a property instead of a function :)

Otherwise everything looks fine

inga-affinidi (changes requested):

common-libs/common/src/services/DidDocumentService/ElemDidDocument.ts
```diff
@@ -1,17 +1,17 @@
 import base64url from 'base64url'
 import { parse } from 'did-resolver'
 
-import { op, func } from './elem-lib'
```
inga-affinidi, line 4: One more thing: you'll need to update the versions in `package.json` and `package-lock.json`; update dependencies in dependent packages; and update changelog.

cosminm-affnd: Thank you :), I always forget this, good target for automation :)

common-libs/common/src/services/DidDocumentService/KeyVault.ts
cosminm-affnd: done

common-libs/common/package-lock.json
```diff
@@ -1,15 +1,17 @@
...
   "requires": true,
   "packages": {
     "": {
       "name": "@affinidi/common",
-      "version": "2.0.0-beta.4",
+      "version": "2.0.0-beta.5",
       "license": "ISC",
       "dependencies": {
         "@affinidi/affinity-metrics-lib": "^0.0.25",
+        "@affinidi/tiny-lds-ecdsa-secp256k1-2019": "^1.1.0",
```
inga-affinidi, line 13: In order for monorepo to work properly, you should only ever do `npm run bootstrap` from the project root directory, and never use `npm install` inside the specific packages directories

common-libs/common/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@affinidi/common",
-  "version": "2.0.0-beta.4",
+  "version": "2.0.0-beta.5",
```
inga-affinidi, line 3: Please also update the references in dependent packages (sdk\core etc).

No need to update dependent packages versions, since behavior is not changed there; just update their `@affinidi/common` dependencies to beta.5. This only affects package.json; package-lock.json does not refer to @affinidi/common.

inga-affinidi (changes requested):

common-libs/common/package-lock.json
cosminm-affnd: reverted the change

common-libs/common/package.json
cosminm-affnd: done

inga-affinidi (dismissed):

DenisPopov15 (dismissed):

DenisPopov15 (approved):

inga-affinidi (approved):

PR #131, opened 2021-08-23
Export LocalKeyVault

cosminm-affnd:
Export LocalKeyVault so it can be used by other libraries.

### DIFF
--- a/common-libs/common/CHANGELOG.md
+++ b/common-libs/common/CHANGELOG.md
@@ -1,4 +1,4 @@
-# 2.0.0-beta.5 (2020-08-23)
+# 2.0.0-beta.6 (2020-08-23)
 * export `LocalKeyService`
 # 2.0.0-beta.5 (2020-08-20)
 * `ElemDidDocument` abstracted to use a KeyVault to provide the relevant keys
--- a/common-libs/common/CHANGELOG.md
+++ b/common-libs/common/CHANGELOG.md
@@ -1,3 +1,5 @@
+# 2.0.0-beta.5 (2020-08-23)
+* export `LocalKeyService`
 # 2.0.0-beta.5 (2020-08-20)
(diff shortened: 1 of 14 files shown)


dendavidov-affinidi (dismissed):

common-libs/common/CHANGELOG.md
```diff
@@ -1,3 +1,5 @@
+# 2.0.0-beta.5 (2020-08-23)
```
DenisPopov15, line 1: @cosminm-affnd should be `2.0.0-beta.6` here ?

DenisPopov15 (dismissed):

common-libs/common/CHANGELOG.md
cosminm-affnd: oups, updated

DenisPopov15 (approved):

common-libs/common/src/services/DidDocumentService/KeyVault.ts
```diff
@@ -23,4 +23,11 @@ export interface KeyVault {
    * @param payload payload to be signed
    */
   sign(payload: Buffer): Buffer
+
+  /**
+   * Sign the given payload with the primary public key. Same as sign but async.
+   *
+   * @param payload payload to be signed
+   */
+  signAsync?(payload: Buffer): Promise<Buffer>
```
Yengas, line 32: I did not get how this helps me. I still need to implement the sync version of this function right. No one uses the `signAsync` as well? 🤔

cosminm-affnd: The issue I had was when implementing a KeyVault backed by metamask, the sign method there is async, and adding an optional async method here was the only work around I could find. Other approaches I could use?

Yengas (approved):

PR #137, opened 2021-08-30
FTL-1707 bbs user creation support

inga-affinidi:
All the code from `affinidi-vc-signatures` moved to SDK (into `@affinidi/common`, `wallet-core-sdk` and `wallet-node-sdk`).

`affinidi-vc-signatures` contained 478 lines of code (not including tests).
With this PR, total amount of code in SDK increased by roughly 300 lines; additionally, total amount of code in `cloud-wallet-api` decreased by roughly 250 lines ([WIP](https://gitlab.com/affinidi/foundational/cloud-wallet-api/-/merge_requests/119))


common-libs/common/src/shared/EcdsaCryptographyTools.ts
```diff
@@ -0,0 +1,51 @@
...
+import { IPlatformCryptographyTools } from './interfaces'
+
+export const ecdsaCryptographyTools: IPlatformCryptographyTools = {
+  signSuites: {
+    bbs: () => {
+      throw new Error('Not implemented')
+    },
+    rsa: () => {
+      throw new Error('Not implemented')
+    },
```
vitalyrudenko-affinidi, line 11: It's a bit confusing that these are here. 
Why not have `signSuite`, `generateKey`, `verifySuiteFactory` here and don't mention bbs/rsa?

What I mean is, imho, we should have the opposite:
```
{
    rsa: { signSuite, generateKey, verifySuiteFactory },
    ecdsa: { signSuite, generateKey, verifySuiteFactory },
    bbs: { signSuite, generateKey, verifySuiteFactory, deriveSegmentProof, validateSegmentProof },
}
```

Therefore we won't have to mention bbs/rsa in this ecdsa helper file.

sdk/core/src/CommonNetworkMember/NetworkMemberWithCognito.ts
```diff
@@ -47,84 +46,56 @@ export abstract class NetworkMemberWithCognito extends BaseNetworkMember {
...
       keyParamsOrOptions,
     )
   }
 
-  /**
-   * @deprecated Temporary implementation; refactor by 6.0 release (FTL-1707)
-   */
   private static _createKeyParamsOrOptionsValidator(keyParamsOrOptions?: KeyParamsOrOptions) {
-    if (!keyParamsOrOptions || 'encryptedSeed' in keyParamsOrOptions) {
+    if (!keyParamsOrOptions || NetworkMemberWithCognito._isKeyParams(keyParamsOrOptions)) {
```
vitalyrudenko-affinidi, line 86: I think in static classes with static methods you can use `this`

vitalyrudenko-affinidi (changes requested):

common-libs/common/src/shared/EcdsaCryptographyTools.ts
inga-affinidi: 1. We mention bbs/rsa in ecdsa helper file because `Affinity.ts` uses these, and the alternative would be to make them optional so that `Affinity.ts` will have to check that the object is not undefined and the function is not undefined prior to calling it, and throw the more cryptic error otherwise. With the current implementation, there are no optional fields, code in `Affinity.ts` does not require any additional checks, and exceptions contain informative traces.

2. Note that `keyGenerators` only has `rsa` and `bbs` fields, while for `verifySuiteFactories` keys are entirely different. It is not just "verify a ecdsa signature", it is specifically "verify signature for credential with `proof.type === 'EcdsaSecp256k1Signature2019'`.

3. And this also makes code simpler in `Affinity.ts`: when it needs to verify a signature, it does not have to concern itself with "what proof type corresponds to what cryptographic function" (and what if there are several proof types relying on the same cryptographic function?), it can just get a verifier for proof type.

sdk/core/src/CommonNetworkMember/NetworkMemberWithCognito.ts
inga-affinidi: `this` meaning is subtly different, and TS will remind you of this as soon as you put any constraints on `this`.

We already have a bunch of these static methods in this class, they have to call other private static methods using class name instead of this (otherwise TS will not build the code).

In this specific case, considering that caller is private as well, and that it is never passed anywhere but only called directly, we could replace explicit class name with `this`, but that's probably not worth it, with all these qualifiers and with other methods using explicit class name anyway.

vitalyrudenko-affinidi (dismissed):

vitalyrudenko-affinidi (approved):

cosminm-affnd (approved):

Yengas (approved):

PR #138, opened 2021-09-01
FTL-1912 fix mutation in GenericApiService

Yengas:
## Context
https://replika.atlassian.net/browse/FTL-1912

## Description
Mutation in GenericApiService causes bundling in some platforms to fail. Better to get rid of it.
image

### DIFF
--- a/common-libs/internal-api-clients/src/services/GenericApiService.ts
+++ b/common-libs/internal-api-clients/src/services/GenericApiService.ts
@@ -71,8 +71,8 @@ export default class GenericApiService<TApi extends BuiltApiType> {
 
     // eslint-disable-next-line no-unused-vars
     const path = pathTemplate.replace(/\{(\w+)\}/g, (_match, p1) => options.pathParams?.[p1])
-    const url = new URL(path)
-    url.search = new URLSearchParams(options.queryParams ?? {}).toString()
+    const search = new URLSearchParams(options.queryParams ?? {}).toString()
+    const url = new URL(path.endsWith('?') || !search ? path + search : `${path}?${search}`)
 
     const response = await fetch(url, fetchOptions)
     const { status } = response
--- a/common-libs/internal-api-clients/src/services/GenericApiService.ts
+++ b/common-libs/internal-api-clients/src/services/GenericApiService.ts
@@ -60,7 +60,7 @@ export default class GenericApiService<TApi extends BuiltApiType> {
     method: string,
     pathTemplate: string,
     headers: ApiRequestHeaders,
-    options: { params?: any; pathParams?: any; queryParams?: any },
+    options: { params?: any; pathParams?: any; queryParams?: Record<string, string> },
   ) {
     const { params } = options
     const fetchOptions = {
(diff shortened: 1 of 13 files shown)


cosminm-affnd (dismissed):

common-libs/internal-api-clients/src/services/GenericApiService.ts
```diff
@@ -71,8 +71,8 @@ export default class GenericApiService<TApi extends BuiltApiType> {
 
     // eslint-disable-next-line no-unused-vars
     const path = pathTemplate.replace(/\{(\w+)\}/g, (_match, p1) => options.pathParams?.[p1])
-    const url = new URL(path)
-    url.search = new URLSearchParams(options.queryParams ?? {}).toString()
+    const search = new URLSearchParams(options.queryParams ?? {}).toString()
+    const url = new URL(path.endsWith('?') || !search ? path + search : `${path}?${search}`)
```
inga-affinidi, line 75: It was not a mutation, it was using URL API to build a new URL instead of manually composing URL string.
AFAIK the setters are there in https://url.spec.whatwg.org/ specifically to allow developers to build URLs from parts instead of manually composing strings.

Are there any other ways to solve this? Maybe `expo` provides some other similar component to build URLs, if its own URL object does not implement the standard?

sdk/core/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@affinidi/wallet-core-sdk",
-  "version": "6.0.0-beta.8",
+  "version": "6.0.0-beta.9",
```
Yengas, line 3: need to update browser/expo/node/rn sdks to use latest core-sdk

common-libs/internal-api-clients/src/services/GenericApiService.ts
inga-affinidi: Also do we have any link describing incompatibility between expo's URL and whatwg's standard URL?

Yengas: maybe we can use a native JS library to compose the URL from parts but IMHO no need to add a new package, or change strategies according to platform just for avoiding string concatenation

DenisPopov15 (dismissed):

common-libs/internal-api-clients/src/services/GenericApiService.ts
vitalyrudenko-affinidi: What was the problem here?

vitalyrudenko-affinidi: How about this?
```js
const url = new URL(path)
for (const [key, value] of Object.entries(options.queryParams)) {
    url.searchParams.set(key, value)
}
```

DenisPopov15 (approved):

vitaly-rudenko (approved):

cosminm-affnd (approved):

DenisPopov15 (approved):

common-libs/internal-api-clients/src/services/GenericApiService.ts
inga-affinidi: @Yengas , an answer to that would probably help us in the future:

> Also do we have any link describing incompatibility between expo's URL and whatwg's standard URL?

PR #140, opened 2021-09-03
feat: Adding business reference credential

hiteshjoshi1:


sdk/vc-data/package.json
```diff
@@ -20,7 +20,7 @@
     "build": "tsdx build --tsconfig ./tsconfig.build.json",
     "test": "tsdx test",
     "test:watch": "tsdx test --watch",
-    "lint": "tsdx lint src",
+    "lint": "tsdx lint src --fix",
```
inga-affinidi, line 23: Why do we need this change? `npm run lint` is executed by CI; we want to make sure there are no linting errors, not to fix what is fixable?

DenisPopov15 (dismissed):

oleksiipiliugin (dismissed):

DenisPopov15 (approved):

PR #141, opened 2021-09-09
[FTL-1925] Cache resolved DIDs

inga-affinidi:
TODO:
- [x] figure out why tests are failing in the packages that were not changed
- [x] bump versions, write changelogs

Changes:

* `SdkError` and `profiler` moved out of `@affinidi/common` to a separate package `@affinidi/tools-common` (you can propose other name if you don't like it) so that `internal-api-clients` will no longer have to depend on `@affinidi/common`.
* `@affinidi/common` now depends on `internal-api-clients` instead of manually sending requests to a registry service.
* `DidResolver` implemented: a drop-in replacement for `RegistryApiService` for did resolving. For every serviceUrl+apiKey combination, it caches promises for every requested did, so that at most one successful request is performed for every did.
* All dids are now resolved using `DidResolver`, this should significantly improve real-world performance, and make integration tests faster.
* Did auth services in `internal-api-clients` now use did auth session objects, making testing them easier (in theory), and allowing in the future to share the same session / did auth token between different services.
* `internal-api-clients` refactored to use composition instead of inheritance, making testing them easier (in theory), and getting read of manually working with TS types in exported services.
* These two points also mean a bit more type safety for did auth services.
* Removed unneeded type recursion from `openapiParser` to make TS compiler life easier.
* Additional improvements to CI performance:
  * Separated `bootstrap` and `build` steps, so that lerna could better parallelize build.
  * Removed `prepare` steps (which just invoked `npm run build` before) to reduce confusion and improve performance of `npm install` / `npm run bootstrap`.
  * Removed lint from lerna build step because we're calling lint anyway later in the pipeline.
  * Removed tsc call from lerna build step because it is calling tsc indirectly by calling build anyway.
  * Configured license checker to only output summary, to improve I/O performance (previously, license checker spent almost half of its runtime on writing detailed report to the output)


common-libs/internal-api-clients/src/services/KeyStorageApiService.ts
```diff
@@ -30,41 +28,43 @@ type GetSignedCredentialRequest = {
...
+
   constructor(options: ConstructorOptions) {
-    super(options.keyStorageUrl, options, keyStorageSpec)
+    this.options = createClientOptions(options.keyStorageUrl, options)
   }
 
-  async storeTemplate(params: ApiType['StoreTemplate']['requestBody']) {
-    return this.execute('StoreTemplate', { params })
+  async storeTemplate(params: GetParams<typeof client.StoreTemplate>) {
+    return client.StoreTemplate(this.options, { params })
```
vitalyrudenko-affinidi, line 44: Let's move client creation to the constructor and pass "this.options" to its constructor instead of repeating this argument?

```
export default class KeyStorageApiService {
  constructor(options: ConstructorOptions) {
    const options = createClientOptions(options.keyStorageUrl, options)
    this.client = createClient(keyStorageSpec, options)
  }

  async readMyKey({ accessToken }: { accessToken: string }) {
    return this.client.ReadMyKey({ authorization: accessToken })
  }
```

common-libs/common/src/shared/DidResolver.ts
```diff
@@ -0,0 +1,57 @@
...
+  private readonly _service: ServiceWithCache
+
+  constructor(options: ConstructorOptions) {
+    this._service = getService(options)
+  }
+
+  resolveDid(did: string) {
+    return resolveDid(this._service, did)
+  }
+}
```
vitalyrudenko-affinidi, line 57: It feels like it's going to be a memory leak

inga-affinidi: It stores:
* For every combination of serviceUrl + accessToken that was ever passed to it: a service instance and a cached set of promises (initially empty)
* For every service instance and every did that was ever passed to it: a promise with the resolved DID.

If some application that uses SDK runs for a long time and resolves a lot of different DIDs, the cache will grow, obviously (because it will have to store all the DIDs it ever resolved in this process). Is that the leak you're talking about?

Was our SDK optimized for that kind of workload anyway?

vitalyrudenko-affinidi: Yeah, exactly. 

> Was our SDK optimized for that kind of workload anyway?

I don't know, but it seems logical to me to expire some values eventually, or have some limit to the cache, because we don't know how our SDK is used - it might as well be used to resolve a lot of DIDs - or it's an edge case?

common-libs/internal-api-clients/src/helpers/client.ts
```diff
@@ -0,0 +1,149 @@
+import keyBy from 'lodash.keyby'
+import mapValues from 'lodash.mapvalues'
+import FetchType from 'node-fetch'
```
vitalyrudenko-affinidi, line 3: Can we do this?
```
import type FetchType from 'node-fetch'
```

Because I suppose this is still an import in the compiled JS code?

common-libs/internal-api-clients/src/helpers/client.ts
```diff
@@ -0,0 +1,149 @@
...
+type GetRequestOptions<TOperation extends MethodTypeByOperation<any>> = Parameters<TOperation>[0]
+export type GetParams<TOperation extends MethodTypeByOperation<any>> = GetRequestOptions<TOperation>['params']
+
+const parseSpec = (rawSpec: GenericApiSpec) => {
+  const basePath = rawSpec.servers[0].url
+
+  const spec = Object.entries(rawSpec.paths).flatMap(([operationPath, operation]) => {
+    const path = `${basePath}${operationPath}`
+
+    return (['get', 'post', 'put', 'delete'] as const).flatMap((method) => {
```
vitalyrudenko-affinidi, line 54: This hardcoded list worries me. If we're going to add some PATCH routes on our API, it won't work here until we update this list?

common-libs/internal-api-clients/src/helpers/didAuthClientWrapper.ts
```diff
@@ -0,0 +1,78 @@
...
+    // so we have to make a temporary object here in order for function to get a display name
+    // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/name#inferred_function_names
+    const functionObject = {
+      [key]: async function (this: ThisData, requestOptions: any) {
+        const authorization = await this.didAuthSession.getResponseToken((did) => createRequestToken(this, did))
+        return method.call(this, { ...requestOptions, authorization })
+      },
+    }
+
+    return functionObject[key]
```
vitalyrudenko-affinidi, line 62: I've seen this in an another file. 

Can we make a util for that so we don't duplicate code & comments?
Something like:
```
return createNamedFunction(key, async () => { ... })
```

inga-affinidi: @vitalyrudenko-affinidi , it would look really cumbersome and add a new entry to the stack trace, so I'm not sure if it's worth it:

```
const createNamedFunction = <TThis, TRest extends any[], TResult>(name: string, f: (self: TThis, ...rest: TRest) => Promise<TResult>)) => {
  const functionObject = {
    [key]: async function (this: TThis, ...parameters: TRest) {
      const result = await f(this, ...parameters)
      return result
    },
  }

  return functionObject[key]
}
```

I'm not sure if it makes sense to do that just for these two places in code.

common-libs/internal-api-clients/src/helpers/client.ts
inga-affinidi: If we're going to add some PATCH routes, build will fail until we add PATCH support because of https://github.com/affinityproject/affinidi-core-sdk/blob/f9b09b397570a903567e0d378c3dc5bb7d5f4db9/common-libs/internal-api-clients/src/types/openapi.ts#L136

(Also this code wasn't changed, before this PR it just resided in another file)

But I can add support for other methods if you feel like we need it here.

common-libs/internal-api-clients/src/helpers/client.ts
inga-affinidi: Yes, we should be able to do this.

common-libs/common/src/shared/DidResolver.ts
inga-affinidi: I think it's an edge case, and implementing an expiration is probably more trouble than would be worth it.

common-libs/internal-api-clients/src/helpers/didAuthClientWrapper.ts
vitalyrudenko-affinidi: Okay, I agree. This util still sounds useful to me, so let's keep it in mind and if it appears in a third place - let's move it.

common-libs/internal-api-clients/src/services/KeyStorageApiService.ts
inga-affinidi: (Fixed)

common-libs/common/src/shared/DidResolver.ts
vitalyrudenko-affinidi: Okay, let's start with the current solution and see if there are any real problems.

common-libs/internal-api-clients/src/helpers/didAuthClientWrapper.ts
inga-affinidi: It might be more useful as a `mapFunctions` alternative to lodash's `mapValues`

common-libs/internal-api-clients/package.json
```diff
@@ -43,8 +42,9 @@
   "homepage": "https://github.com/affinityproject/affinidi-core-sdk#readme",
   "dependencies": {
     "@affinidi/affinidi-did-auth-lib": "^1.0.8",
```
Yengas, line 44: can we make the version 1.0.8 until we solve the issues with the did auth lib versioning?

anton-iskryzhytskyi:
Good work!
According to name of the PR, it should add only caching of DIDs but also contains a lot of others that not related (e.g. CI pipelines, refactorings, improvements.  
IMHO it is easier for reviewers to get into a small amount of changes related to one topic. So my question is, what do you think about splitting this PR into several small:  e.g.:

common-libs/internal-api-clients/src/services/RevocationApiService.ts
```diff
@@ -1,36 +1,40 @@
...
 type ConstructorOptions = DidAuthConstructorOptions & { revocationUrl: string }
 
-type ApiType = BuildApiType<ParseSpec<typeof revocationSpec>>
-
 type ReplaceFieldsWithAny<T> = {
   [key in keyof T]: any
 }
 
+const { CreateDidAuthRequest, ...otherMethods } = createClientMethods(revocationSpec)
+const clientMethods = wrapWithDidAuth(CreateDidAuthRequest, otherMethods)
```
Yengas, line 15: when we are wrapping the client methods with did auth; can we enable/disable did auth for specific endpoints?

for example revocation has an endpoint that can be used without did-auth not usd in SDK. But if we were to add that endpoint to this service class, how could we "not have did auth" for that specific endpoint?

with the old code you could either call;
```
this.execute
this.executeWithDidAuth
```

common-libs/common/src/shared/DidResolver.ts
```diff
@@ -0,0 +1,57 @@
...
+
+const services: Map<string, ServiceWithCache> = new Map()
+
+const getService = (options: ConstructorOptions) => {
+  const { accessApiKey, registryUrl } = options
+  const cacheKey = JSON.stringify({ accessApiKey, registryUrl })
+  if (!services.has(cacheKey)) {
+    services.set(cacheKey, {
+      service: new RegistryApiService(options),
+      cache: new Map(),
```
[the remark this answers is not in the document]
inga-affinidi: We already have a limited version of this in `Affinidi`.

If needed, we could also make a wrapper around `DidResolver` with a pre-populated cache. We cannot do this in `DidResolver` itself because pre-populated cache cannot be trusted and therefore cannot be reused between users (and `DidResolver` _is_ reused between users)

common-libs/internal-api-clients/src/services/RevocationApiService.ts
inga-affinidi:
```
const { CreateDidAuthRequest, MethodWithoutDidAuth1, MethodWithoutDidAuth2, ...methodsWithDidAuth } = createClientMethods(spec)
const wrappedMethodsWithAuth = wrapWithDidAuth(CreateDidAuthRequest, otherMethods)

class Service {
  private client
  private authenticatedClient

  constructor(...) {
    this.client = createClient({ MethodWithoutDidAuth1, MethodWithoutDidAuth2 }, ...)
    this.authenticatedClient = createClientWithAuth(wrappedMethodsWithAuth, didAuthSession, ...)
  }

  method1(...) {
    return this.client.MethodWithoutDidAuth1(...)
  }

  method3(...) {
    return this.authenticatedClient.MethodWithDidAuth3(...)
  }
}
```

inga-affinidi: But the better option would probably to have separate service classes for authenticated and non-authenticated methods.

Which would also allow to call non-authenticated methods without having to supply a correct did / DidAuthAdapter (which would not be used anyway) to the service constructor.

common-libs/internal-api-clients/package.json
inga-affinidi: Will this solve the issue?

inga-affinidi: Fixed version at 1.0.8

common-libs/internal-api-clients/src/helpers/didAuthClientWrapper.ts
inga-affinidi: Implemented `mapFunction` but not sure that the code became more maintainable or clear.

common-libs/internal-api-clients/src/helpers/client.ts
inga-affinidi: Fixed

vitalyrudenko-affinidi (approved):

vitalyrudenko-affinidi (approved):

Yengas (approved):

PR #144, opened 2021-09-14
FTL-1670 add support of elem-anchored method

anton-iskryzhytskyi:
TODO:
- [x] - update sdk to allow elem-anchored
- [x] - update sdk tests to check elem-anchored related 
- [ ] - bump versions, write changelogs

Notes:
* all changes based on this closed [PR](https://github.com/affinityproject/affinidi-core-sdk/pull/124)

Changes:
* add new did method - elem-anchored
* move the work with registry-api from central class Affinity to separate service
* update seed parsing to allow storing metadata
* implement ElemAnchoredDidDocument service


inga-affinidi:
Please postpone this until #141 is merged, it should address some of the issues addressed there.

common-libs/common/src/services/DidDocumentService/ElemAnchoredDidDocument.ts
```diff
@@ -0,0 +1,26 @@
+import KeysService from '../KeysService'
+import { DidResolver } from '../../shared/DidResolver'
+
+export default class ElemDidDocument {
+  private readonly _keysService: KeysService
+  private readonly _signingKey: string
+  private readonly _didResolver: DidResolver
+
+  constructor(keysService: KeysService, didRosolver: DidResolver) {
```
inga-affinidi, line 9: `didResolver`

common-libs/common/src/services/DidDocumentService/index.ts
```diff
@@ -1,28 +1,24 @@
...
-        didDocumentService = new ElemDidDocument(new LocalKeyVault(keysService))
-        break
-    }
-
-    return didDocumentService
+    return {
+      jolo: new JoloDidDocument(keysService),
+      elem: new ElemDidDocument(new LocalKeyVault(keysService)),
+      'elem-anchored': new ElemAnchoredDidDocument(keysService, didResolver),
+    }[didMethod]
```
inga-affinidi, line 21: Seems like a nice opportunity to also refactor this constructor into a factory (and mark constructor as deprecated, for removal in SDK v7).

Constructors shouldn't return values :)

common-libs/common/src/services/RegistryResolveDidService.ts
```diff
@@ -0,0 +1,33 @@
...
+
+if (!fetch) {
+  fetch = require('node-fetch')
+}
+
+export class RegistryResolveDidService {
+  constructor(private readonly _registryUrl: string, private readonly _apiKey: string) {}
+
+  async resolveDid(did: string): Promise<any> {
+    const url = `${this._registryUrl}${RESOLVE_DID_PATH}`
```
inga-affinidi, line 13: Please use `RegistryApiService` instead

inga-affinidi: or higher-level `DidResolver`, which seems to be more appropriate here

common-libs/common/src/services/RegistryResolveDidService.ts
```diff
@@ -0,0 +1,33 @@
+const RESOLVE_DID_PATH = '/api/v1/did/resolve-did'
+
+let fetch: any
+
+if (!fetch) {
+  fetch = require('node-fetch')
+}
+
+export class RegistryResolveDidService {
```
inga-affinidi, line 9: Although it doesn't seem to be used anywhere?

common-libs/common/src/shared/seedTools.ts
```diff
@@ -10,35 +10,46 @@ export type KeyOptions = {
...
-  const { seedHex, didMethod, base64EncodedKeys } = getDecryptedSeedParts(decryptedSeed)
+export const parseDecryptedSeed = (decryptedSeed: string): ParseDecryptedSeedResult => {
+  const { seedHex, didMethod, base64EncodedKeys, base64EncodedMetadata } = getDecryptedSeedParts(decryptedSeed)
 
   validateDidMethodSupported(didMethod)
   const seedHexWithMethod = `${seedHex}++${didMethod}`
   const seed = Buffer.from(seedHex, 'hex')
+  let fullSeedHex = seedHexWithMethod
+  let externalKeys
+  let metadata
```
inga-affinidi, line 40: please avoid `let` usage if possible, it makes code harder to follow

common-libs/common/src/shared/seedTools.ts
```diff
@@ -10,35 +10,46 @@ export type KeyOptions = {
...
-    }
+  if (base64EncodedKeys) {
+    fullSeedHex = `${fullSeedHex}++${base64EncodedKeys}`
+    externalKeys = JSON.parse(base64url.decode(base64EncodedKeys))
   }
 
-  const fullSeedHex = `${seedHexWithMethod}++${base64EncodedKeys}`
-  const externalKeys = JSON.parse(base64url.decode(base64EncodedKeys))
+  if (base64EncodedMetadata) {
+    fullSeedHex = `${fullSeedHex}++${base64EncodedKeys ? '' : '++'}${base64EncodedMetadata}`
```
inga-affinidi, line 48: This is _really_ hard to follow, in particularly because `fullSeedHex` is mutable.

It goes like this:
1. By default, `fullSeedHex` is `seedHexWithMethod`.
2. But if there are keys, we append `++${base64EncodedKeys}` to it.
3. But if there is a metadata, we append `++` to it, and another `++` if there were no keys, and then `base64EncodedMetadata`.
4. And instead of appending things by `fullSeedHex += ...` in the previous two steps, we append them by
    ```
    fullSeedHex = `${fullSeedHex}...`
    ```

Why not just do
```
const fullSeedHex = `${seedHexWithMethod}++${base64EncodedKeys}++${base64EncodedMetadata}`
```
instead, without any conditions?
That way it will be clear what fullSeedHex should look like, and what do different parts mean.

common-libs/common/test/unit/services/DidDocumentService.test.ts
```diff
@@ -28,6 +28,14 @@ const encryptedSeedElem =
...
   'f6d18b619e97a2033f0ec7a6630fdcd3827a0dd70b3c439ab4' +
   '76b3fc264b639c84935d6e5d6fcabb3d027d411ae5d74d570fd16d604b038a9250ce4ac271fd6a86d8401ac52c52'
+const encryptedSeedElemAnchored =
+  '58a387ea33ffa071707a4e3fdf4e9042f11e8512fcc51193b066' +
+  '73c3f7b6c7edf7acdb2ee70eda9c69f39640b472cb38c78bb661dd' +
+  '7bdf1cefbf5eefb30625a3eb6b64776238cf8585edea5410f956566cb0' +
+  '16f6a567ee9561801c0f12b445aa5c9255aa6c68842bb82ae1c7a6c823c0a' +
+  '4f6c4f50a420e18c4513b226e798d9bfbd4ec746f45a0d3a7b43e09ac1348332' +
+  'a3a54748a0f8b5ea917c273f7303be545dde3b95ac6dc5ded4f548fca00bbcc5e8' +
+  '00a61b0f36885ce50e986370045c190f6054c0bf1992740ebaf2874eb7f6f'
```
inga-affinidi, line 38: Is there any logic to it being splitted into lines in this way and not another?

common-libs/common/src/shared/seedTools.ts
```diff
@@ -10,35 +10,46 @@ export type KeyOptions = {
...
+  externalKeys?: any[]
+  metadata?: Record<string, any>
+}
+
 export const getDecryptedSeedParts = (decryptedSeed: string) => {
   const seedParts = decryptedSeed.split('++')
   const seedHex = seedParts[0]
   const didMethod = seedParts[1]
   const base64EncodedKeys = seedParts[2]
+  const base64EncodedMetadata = seedParts[3]
```
inga-affinidi, line 27: Now I see there was a bug in original implementation, and this bug continues into this PR as well.

Unfortunately, since keys are base64-encoded, they can contain `++` substrings (as opposed to `seedHex` and `didMethod`).
And if we just split `seedParts` by `++`, we'll end up splitting these keys.

We should use something else as a separator beyond didMethod, and probably also do splitting manually.

inga-affinidi: Also I would suggest to put an end to this expansion of decryptedSeed format, and do something like:
```
`${seedBinary}` // extremely old format
`${seedHex}++${didMethod}` // old format
`${seedHex}++${didMethod}++${base64EncodedKeys, can contain '++'}` // old format, only appeared in SDK v6 beta, so not used much
`${seedHex}++${didMethod};additionalData=${base64EncodedData}` // final format, contains both keys and metadata and all the other data we might want to add in the future
```

common-libs/common/src/services/DidDocumentService/index.ts
anton-iskryzhytskyi: yes, you are right, I also thought about that. let's create a doc with possible refactorings and then when we have a time - we will work on that

common-libs/common/src/services/RegistryResolveDidService.ts
anton-iskryzhytskyi: should be removed

common-libs/common/src/services/RegistryResolveDidService.ts
anton-iskryzhytskyi: yes, removed=)

common-libs/common/test/unit/services/DidDocumentService.test.ts
anton-iskryzhytskyi: Good question, related to prettier I think. Is it make important  from your side?

common-libs/common/src/shared/seedTools.ts
anton-iskryzhytskyi: `${seedHex}++${didMethod}++;additionalData:${base64EncodedData}`
so, base64EncodedData this also should contains base64EncodedKeys as well with meta?

common-libs/common/src/services/DidDocumentService/index.ts
inga-affinidi: It is a really small refactoring and it makes sense within the context of this PR, so I'd suggest to do it here.

common-libs/common/test/unit/services/DidDocumentService.test.ts
inga-affinidi: Prettier would have probably suggested to split the large line into equal-length chunks (probably 120 characters long), but here all chunks are of wildly different lengths (58, 60, 64, 67, 70, 72 and 65). Is it meaningful?

common-libs/common/src/shared/seedTools.ts
inga-affinidi: Yes, base64EncodedData can contain JSON with `keys` and `metadata`.

Of course this will make seed slightly larger (because we'll base64 encode keys that are already base64-encoded).

Or maybe if that is a concern we could use BSON instead of JSON.

common-libs/common/test/unit/services/DidDocumentService.test.ts
anton-iskryzhytskyi: Anyway I'll change the elemAnchoredSeed, so copy existing approach

common-libs/common/src/shared/seedTools.ts
anton-iskryzhytskyi: It could be an option, let's try this format now

common-libs/common/src/services/DidDocumentService/index.ts
anton-iskryzhytskyi: ok

common-libs/common/src/services/DidDocumentService/ElemAnchoredDidDocument.ts
```diff
@@ -0,0 +1,26 @@
+import KeysService from '../KeysService'
+import { DidResolver } from '../../shared/DidResolver'
+
+export default class ElemDidDocument {
```
inga-affinidi, line 4: Shouldn't it be `ElemAnchoredDidDocument`?

(Also we probably need to configure an eslint rule for that...)

common-libs/common/src/services/DidDocumentService/index.ts
```diff
@@ -1,28 +1,31 @@
...
 
-    return didDocumentService
+    return {
+      jolo: new JoloDidDocument(keysService),
+      elem: new ElemDidDocument(new LocalKeyVault(keysService)),
+      'elem-anchored': new ElemAnchoredDidDocument(keysService, didResolver),
+    }[didMethod]
   }
 
   getMyDid() {
```
inga-affinidi, line 31: So all these non-static methods are not needed anymore, because there never were any instances of `DidDocumentService`, and now there are no variables with that type as well.

common-libs/common/src/services/DidDocumentService/ElemAnchoredDidDocument.ts
```diff
@@ -0,0 +1,26 @@
...
+  private readonly _didResolver: DidResolver
+
+  constructor(keysService: KeysService, didResolver: DidResolver) {
+    this._signingKey = 'primary'
+    this._keysService = keysService
+    this._didResolver = didResolver
+  }
+
+  getMyDid(): string {
+    const { anchoredDid } = this._keysService.getMetadata()
```
inga-affinidi, line 16: And `getMetadata` decrypts a seed on every call. Is it really a good idea to do that whenever we need to obtain DID?

Maybe it would be better to use KeyVault here instead of KeysService?

anton-iskryzhytskyi: Not sure, I'm new to this codebase, but as I found, there are a lot of invocations of decryptSeed method
Also, not sure how often it will be invoked, maybe even less than other methods that uses decryptSeed (methods in KeysService).

If you really know that it will affect performance in some places, please help me find solution, in other case, let's use this rule:   
“The real problem is that programmers have spent far too much time worrying about efficiency in the wrong places and at the wrong times; premature optimization is the root of all evil (or at least most of it) in programming.” - The Art of Computer Programming,  Donald Knuth

inga-affinidi: @anton-iskryzhytskyi , I'm not talking about premature optimization here, but we have a higher-level `KeyVault` already, and it seems to be a better place for that kind of functionality than `KeysService`.

Additionally, do we know how this should interact with other kinds of `KeyVault`? If we need this to work with W3 wallet properly, we might need to use a generic `KeyVault` anyway.

common-libs/common/src/shared/seedTools.ts
```diff
@@ -10,74 +10,96 @@ export type KeyOptions = {
...
-  const { seedHex, didMethod, base64EncodedKeys } = getDecryptedSeedParts(decryptedSeed)
+export const parseDecryptedSeed = (decryptedSeed: string): ParseDecryptedSeedResult => {
+  const { seedHex, didMethod, base64EncodedKeys, base64EncodedAdditionalData } = getDecryptedSeedParts(decryptedSeed)
 
   validateDidMethodSupported(didMethod)
   const seedHexWithMethod = `${seedHex}++${didMethod}`
   const seed = Buffer.from(seedHex, 'hex')
+  const additionalData = base64EncodedAdditionalData && JSON.parse(base64url.decode(base64EncodedAdditionalData))
+  const parsedFromBase64ExternalKeys = base64EncodedKeys && JSON.parse(base64url.decode(base64EncodedKeys))
+  const externalKeys = additionalData?.[EXTERNAL_KEYS_KEY] || parsedFromBase64ExternalKeys
```
inga-affinidi, line 55: by the way `??` would be more appropriate than `||` here.

common-libs/common/src/shared/seedTools.ts
```diff
@@ -10,74 +10,96 @@ export type KeyOptions = {
...
-      seedHexWithMethod,
-      fullSeedHex: seedHexWithMethod,
-    }
-  }
-
-  const fullSeedHex = `${seedHexWithMethod}++${base64EncodedKeys}`
-  const externalKeys = JSON.parse(base64url.decode(base64EncodedKeys))
-
-  return { seed, didMethod, seedHexWithMethod, externalKeys, fullSeedHex }
+  return { seed, didMethod, seedHexWithMethod, externalKeys, fullSeedHex: decryptedSeed, metadata }
```
inga-affinidi, line 58: `fullSeedHex` was used to normalize the seed when re-encrypting and re-saving it prior to this PR.

The oldest kind of seeds (that was not even `hex`, it was just a plain `seed` buffer, not even a string) is now handled outside of `seedTools`, but still we might want to normalize seeds here.

(This comment is not a call to action but just some context that can be useful to those who will look at this PR in the future)

common-libs/common/src/shared/seedTools.ts
```diff
@@ -10,74 +10,96 @@ export type KeyOptions = {
...
+export type ParseDecryptedSeedResult = {
+  seed: Buffer
+  didMethod: string
+  seedHexWithMethod: string
+  fullSeedHex: string
+  externalKeys?: any[]
+  metadata?: Record<string, any>
+}
+
+const ADDITIONAL_DATA_SEPARATOR = '++;additionalData:'
```
inga-affinidi, line 22: Let's change it to `++;additionalDataJson:`, just to make it more future-proof in case we'll ever want to change its format.

anton-iskryzhytskyi:
@inga-affinidi good example why even small refactoring could be a big evil in feature requests, I spent a couple of hours (or even more on that) without any results, because methods of DidDocumentService is used in tests and mocks (sinon.stub on DidDocumentService.prototype, better to replace background using of deps with passing them according to SOLIDependencyInjection). It should be refactored, but in a separate story. Due to improvement can't be bigger than the main intention of PR. 
As I know, SideTree future is really important and that is why I concentrated only on that, without side work

inga-affinidi:
> because methods of DidDocumentService is used in tests and mocks (sinon.stub on DidDocumentService.prototype

But they could not have been used anywhere because no classes extended `DidDocumentService` and no instances of `DidDocumentService` could exist?

So simply removing these mocks should not have affected anything (since the mocked methods are never called, and `DidDocumentService.prototype` is never used). If you tried removing them, what problems did it cause?

anton-iskryzhytskyi:
> If you tried removing mocks, what problems did it cause?

It causes making changes that must be a separate PR. 
It is [not all](https://github.com/affinityproject/affinidi-core-sdk/pull/144/commits/c0bf6190c0bcb247721d75dfa3c52b0ec6cb04e2) what should be fixed in tests.

inga-affinidi:
@anton-iskryzhytskyi , I don't understand.

`DidDocumentService` was never used as a class (at least for a very long time). Its constructor was not actually a constructor but a function returning something else. These instance methods were probably only there (that's my guess, this code was there long before me) so that TS would not complain when someone would try to call them on a returned value because TS had no idea `DidDocumentService` constructor is not actually a constructor and does not return a `DidDocumentService` instance (which could not exist).

So since we have refactored it there anyway and finally created a proper static method instead of continuing to use that constructor-which-is-not-a-constructor, all these instance methods (which are and were never called by any code and do not return any reasonable data) can be removed. And in case some test code happens to mock any of these, these mocks can also be safely removed, without making any other changes to the tests code, because these mocks were not required and never affected anything in the first place (they mocked dead code that could never be called).

And I'm not sure what is the problem here? Why can't we just remove these methods and the relevant mocks as a part of this PR? You have already removed the methods. I've tried to find all mocks and only found two lines in `sdk\core\test\unit\CommonNetworkMember.test.ts` (lines 135:136).
Am I missing something?
Do we really have this discussion over removal of two unused lines, removal of which was made possible by the other changes you did in this PR anyway?

I agree that large refactorings need their own JIRA tickets and should be separated from the other changes.
But do we really need to create a new JIRA ticket, groom it, plan it into an iteration, create a separate PR for it, get two separate approvals just to remove a bit of _dead code_ (three fake instance methods in `DidDocumentService` and two lines mocking these methods in tests), removal of which was made possible by your changes in _this_ PR?

anton-iskryzhytskyi:
yes, I've already shared [a commit](https://github.com/affinityproject/affinidi-core-sdk/pull/144/commits/c0bf6190c0bcb247721d75dfa3c52b0ec6cb04e2) which contains **part** of the changes, not all. (e.g.: [filed pipeline](https://github.com/affinityproject/affinidi-core-sdk/runs/3611158921)). After I complete with this important feature request, I'll probably create a PR with changes including in that commit and other possible required changes,

inga-affinidi:
@anton-iskryzhytskyi , I see now what the problem is.

In this PR, you've created a static factory, marked `DidDocumentService` (which was incorrect all this time) as deprecated and turned it into a wrapper for the factory.

Then you've went through all references to `new DidDocumentService` and replaced them with the call to the factory.

However, you only did this for `src` directory, missing `test` directory. You should also do this for `test` directory as well.
The tests were failing not because they should use dependency injection; the tests were failing because you didn't replace `new DidDocumentService` (incorrect) with `DidDocumentService.createDidDocumentService` (correct) in tests as well.

Replacing calls to old `new DidDocumentService` with new `DidDocumentService.createDidDocumentService` in `tests` (as you have already done with `src`) should be a part of this PR.

And after that is done, you can safely remove these three dead methods and two attempts to mock these dead methods.

common-libs/common/src/services/DidDocumentService/ElemAnchoredDidDocument.ts
inga-affinidi: Still no answer for:

> Additionally, do we know how this should interact with other kinds of KeyVault? If we need this to work with W3 wallet properly, we might need to use a generic KeyVault anyway.

common-libs/common/src/services/DidDocumentService/index.ts
```diff
@@ -1,46 +1,34 @@
...
-      did = this.getMyDid()
-    }
-
-    const signingKey = 'primary'
-
-    return `${did}#${signingKey}`
-  }
+  static createDidDocumentService(
+    keysService: KeysService,
+    didResolver?: DidResolver,
```
inga-affinidi, line 23: If `didResolver` is optional here, shouldn't it throw an exception when `didMethod` is `'elem-anchored'` (instead of failing down the road)?

inga-affinidi: OK, I see now that we shouldn't, so probably some comment is needed here to clarify that `didResolver` is only needed when we are trying to `buildDidDocument`.

Or maybe the better idea would be to remove this parameter from constructor entirely, and make `didResolver` a parameter for `buildDidDocument` method. That way it will reduce the opportunity to make bugs by calling `createDidDocumentService` without a second argument (how it is already done in some places) and then calling `buildDidDocument` on a result.

anton-iskryzhytskyi: hmm, if we will pass it as an argument, it means we need to **update** interface of the method or **detect** which type of DidDocumentService is used - not the best way from my side. 
I expect it won't make any issues in future if we will follow good test coverage, even more.

common-libs/common/src/shared/seedTools.ts
inga-affinidi: That this comment is not a call to action does not mean that it should be hidden.

Github allows us to merge PRs with unresolved conversations.

sdk/browser/test/factory/didFactory.ts
```diff
@@ -82,7 +82,7 @@ export const generateTestDIDs = async (): Promise<{
       encryptedSeed: joloEncryptedSeed,
       seedHex: joloSeedHex,
       did: joloDid,
-      didDocument: joloDidDocument,
+      didDocument: joloDidDocument as any,
```
inga-affinidi, line 85: Why do we need `as any` here?

sdk/core/src/CommonNetworkMember/BaseNetworkMember/AnchoringService.ts
```diff
@@ -0,0 +1,80 @@
...
+    private readonly encryptedSeed: string,
+    private readonly password: string,
+    private readonly didDocument: any,
+    private readonly nonce: number,
+    private readonly options: ParsedOptions,
+    private readonly anchoredDidElem?: boolean,
+  ) {}
+
+  private get api() {
+    return new RegistryApiService({
```
inga-affinidi, line 18: Why not create it in constructor, as in all other services?

sdk/core/src/CommonNetworkMember/BaseNetworkMember/AnchoringService.ts
```diff
@@ -0,0 +1,80 @@
...
+  }
+
+  private async anchorDidWithElemMethod() {
+    const did = this.didDocument.id
+    try {
+      await this.api.anchorDid({
+        did,
+        didDocumentAddress: '',
+        ethereumPublicKeyHex: '',
+        transactionSignatureJson: '',
```
inga-affinidi, line 58: Why do we rely on empty strings instead of e.g. null or undefined?

sdk/core/src/CommonNetworkMember/BaseNetworkMember/AnchoringService.ts
```diff
@@ -0,0 +1,80 @@
...
+        anchoredDidElem: this.anchoredDidElem,
+      })
+    } catch (error) {
+      console.log('to check logs at the backend', error)
+    }
+  }
+
+  public async anchorDid() {
+    const keysService = new KeysService(this.encryptedSeed, this.password)
```
inga-affinidi, line 67: But we can just create it in constructor and use it as `this.keysService` in `anchorDidWithJoloMethod`, instead of passing it as a parameter for some reason?

sdk/core/src/CommonNetworkMember/BaseNetworkMember/AnchoringService.ts
```diff
@@ -0,0 +1,80 @@
...
+    const { seed, didMethod } = keysService.decryptSeed()
+    const seedHex = seed.toString('hex')
+
+    if (didMethod === JOLO_DID_METHOD) {
+      return this.anchorDidWithJoloMethod(keysService, seedHex)
+    }
+
+    // NOTE: if anchoredDidElem = false - for metrics purpose only
```
inga-affinidi, line 75: I do not understand this comment. What does it mean, how is it related to `if (didMethod === ELEM_DID_METHOD)`?

inga-affinidi (changes requested):

sdk/core/src/CommonNetworkMember/BaseNetworkMember/AnchoringService.ts
```diff
@@ -0,0 +1,80 @@
...
+import { KeysService } from '@affinidi/common'
+
+export class AnchoringService {
+  constructor(
+    private readonly encryptedSeed: string,
+    private readonly password: string,
+    private readonly didDocument: any,
+    private readonly nonce: number,
+    private readonly options: ParsedOptions,
+    private readonly anchoredDidElem?: boolean,
```
inga-affinidi, line 14: But `anchoredDidElem` doesn't seem to be used anywhere?

sdk/core/src/CommonNetworkMember/BaseNetworkMember/AnchoringService.ts
```diff
@@ -0,0 +1,80 @@
...
+import { RegistryApiService } from '@affinidi/internal-api-clients'
+import { KeysService } from '@affinidi/common'
+
+export class AnchoringService {
+  constructor(
+    private readonly encryptedSeed: string,
+    private readonly password: string,
+    private readonly didDocument: any,
+    private readonly nonce: number,
+    private readonly options: ParsedOptions,
```
inga-affinidi, line 13: We don't need the entire `ParsedOptions` object here, we only need parameters for the `RegistryApiService` constructor.

Or we could just accept an instance of registry api service here.

sdk/core/src/CommonNetworkMember/BaseNetworkMember/AnchoringService.ts
```diff
@@ -0,0 +1,80 @@
+import { ParsedOptions } from '../../shared/getOptionsFromEnvironment'
+import { extractSDKVersion } from '../../_helpers'
+import { ELEM_DID_METHOD, JOLO_DID_METHOD } from '../../_defaultConfig'
+import { RegistryApiService } from '@affinidi/internal-api-clients'
+import { KeysService } from '@affinidi/common'
+
+export class AnchoringService {
```
inga-affinidi, line 7: It is not a part of `BaseNetworkMember` class hierarchy, and should belong to `services` directory.

sdk/core/src/CommonNetworkMember/BaseNetworkMember/BaseNetworkMember.ts
```diff
@@ -1,11 +1,4 @@
-import {
-  DidDocumentService,
-  JwtService,
-  KeysService,
-  MetricsService,
-  Affinity,
-  generateFullSeed,
-} from '@affinidi/common'
+import { DidDocumentService, DidResolver, JwtService, KeysService, MetricsService, Affinity } from '@affinidi/common'
```
inga-affinidi, line 1: What is the point of moving this file? `CommonNetworkMember` directory was created specifically for `BaseNetworkMember` and its descendants.

sdk/core/src/CommonNetworkMember/BaseNetworkMember/RegisteringService.ts
```diff
@@ -0,0 +1,79 @@
+import { ParsedOptions } from '../../shared/getOptionsFromEnvironment'
```
inga-affinidi, line 1: It belongs to the `sdk/core/src/services` directory.

sdk/core/src/CommonNetworkMember/BaseNetworkMember/RegisteringService.ts
```diff
@@ -0,0 +1,79 @@
...
+      this.platformCryptographyTools,
+      this.didMethod,
+      this.keyOptions,
+      {
+        anchoredDid: did,
+      },
+    )
+    const originalSeed = originalFullSeed.split('++')[0]
+    const draftSeed = elemAnchoredSeedDraft.split('++')[0]
+    const elemAnchoredSeed = elemAnchoredSeedDraft.replace(draftSeed, originalSeed)
```
inga-affinidi, line 54: We had a method `generateFullSeed` that was useful because we never needed to generate its separate parts. This method simply generates all the parts and concatenates them together.

Now we need to create a seed that would reuse an existing first part.

The solution there would be to split `generateFullSeed` method in two (e.g. one that creates a full seed from the given first part and params; and another that generates the first part and then invokes the first method; the second method would be equivalent to the existing `generateFullSeed`), and to invoke the first method here.

Not to invoke `generateFullSeed` with unneeded data, and then do some magic with string splitting and string replacement relying on the knowledge of seed internal composition.

Additionally there is `KeysService`, there is `parseDecryptedSeed` method etc; you don't have to manually split strings in order to get the original first part here.

sdk/browser/test/factory/didFactory.ts
anton-iskryzhytskyi: tests failed

sdk/core/src/CommonNetworkMember/BaseNetworkMember/AnchoringService.ts
anton-iskryzhytskyi: doesn't make any sense

sdk/core/src/CommonNetworkMember/BaseNetworkMember/AnchoringService.ts
anton-iskryzhytskyi: don't know

inga-affinidi:
There is a commit "feat: FTL-1670 add support of elem anchoring" made yesterday. However, you marked this PR as "ready for review" five days ago.

So is it ready for review or not? (Apparently it was not ready for review at least until yesterday?)

Is FTL-1670 now implemented entirely, or will there be further commits?

sdk/core/src/CommonNetworkMember/BaseNetworkMember/AnchoringService.ts
inga-affinidi: How does it not make any sense? It's the pattern we've been following throughout the entire codebase.

common-libs/common/src/shared/seedTools.ts
```diff
@@ -10,74 +10,96 @@ export type KeyOptions = {
...
   const didMethod = seedParts[1]
-  const base64EncodedKeys = seedParts[2]
-
-  return { seedHex, didMethod, base64EncodedKeys }
+  const additionalDataSeparatorIndex = decryptedSeed.indexOf(ADDITIONAL_DATA_SEPARATOR)
+  const additionalDataIndex =
+    additionalDataSeparatorIndex > 0 && additionalDataSeparatorIndex + ADDITIONAL_DATA_SEPARATOR.length
+  const base64EncodedAdditionalData = additionalDataIndex && decryptedSeed.slice(additionalDataIndex)
+  // backwards compatibility
+  const base64EncodedKeys = !additionalDataIndex ? seedParts[2] : undefined
```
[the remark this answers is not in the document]
inga-affinidi: Unresolving so that it is visible to those who will look at this PR in the future

sdk/core/src/CommonNetworkMember/BaseNetworkMember/AnchoringService.ts
anton-iskryzhytskyi: image

sdk/core/src/CommonNetworkMember/BaseNetworkMember/AnchoringService.ts
anton-iskryzhytskyi: image
it is a copy of previous comment + condition when it is used for anchoring

sdk/core/src/CommonNetworkMember/BaseNetworkMember/AnchoringService.ts
anton-iskryzhytskyi: used in anchorDidWithElemMethod

sdk/core/src/CommonNetworkMember/BaseNetworkMember/AnchoringService.ts
anton-iskryzhytskyi: image
copy of old code, it is huge and has a lot of things to be refactored, I feel you want to have ideal code here, not sure if we can complete this PR with such review and previous state of the code base

sdk/core/src/CommonNetworkMember/BaseNetworkMember/AnchoringService.ts
anton-iskryzhytskyi: Actually just splitting of previous codebase, not sure where it should be as we will split this ugly class (1100 lines of code) in future to parts, so from my side it makes sense to keep it in one place. That is why I created a subdir in CommonNetworkMemeber for BaseNetworkMemeber

sdk/core/src/CommonNetworkMember/BaseNetworkMember/BaseNetworkMember.ts
anton-iskryzhytskyi: Due to I ope in the future we can refactor this class with 1100 lines of code, so it makes sense to keep all it parts in one place

sdk/core/src/CommonNetworkMember/BaseNetworkMember/RegisteringService.ts
anton-iskryzhytskyi: let's define it as a next refactoring, anyway elem anchored future requires more work and probably we don't have time to refactor all codebase here

sdk/core/src/CommonNetworkMember/BaseNetworkMember/AnchoringService.ts
anton-iskryzhytskyi: but anyway, makes sense for me

sdk/core/src/CommonNetworkMember/BaseNetworkMember/AnchoringService.ts
anton-iskryzhytskyi: make sense

sdk/core/src/CommonNetworkMember/BaseNetworkMember/RegisteringService.ts
anton-iskryzhytskyi: ^^
Due to I ope in the future we can refactor this class with 1100 lines of code, so it makes sense to keep all it parts in one place

sdk/core/src/CommonNetworkMember/BaseNetworkMember/AnchoringService.ts
inga-affinidi: Originally there was an ugly class called `CommonNetworkMember` with 3000 lines, it was split into smaller interdependent parts all put into `CommonNetworkMember` directory.

It's great that you decided to extract some more independent parts out of it in this PR. It does not make any sense to store these parts in `CommonNetworkMember` directory; they are independent after all. It does not make any sense at all to move `BaseNetworkMember` in its separate directory inside `CommonNetworkMember`.

sdk/core/src/CommonNetworkMember/BaseNetworkMember/RegisteringService.ts
inga-affinidi: I do not agree that we should write a bunch of unmaintainable, unclear and brittle code in this PR and refactor it at some time later in the future, instead of writing maintainable code (which would also be 3x smaller) from the beginning.

CC @carlos-affinidi

carlos-affinidi: Having Sidetree is indeed a priority and understandably @anton-iskryzhytskyi is pushing some refactoring to the backlog. @anton-iskryzhytskyi doesn't have the same knowledge on the SDK that you have @inga-affinidi. How can we enable developers to more easily contribute on the SDK? Do we need a more in depth documentation explaining the different parts and intricacies of the SDK? @inga-affinidi please add these guidelines to the SSDLC document, so we write maintainable code from the beginning as much as possible.

Also, if you really consider that this refactor should be done right away, let's find a way to help @anton-iskryzhytskyi with them so he can focus on Sidetree. Let's not block him at this point, even if that means a new MR.

carlos-affinidi:
> There is a commit "feat: FTL-1670 add support of elem anchoring" made yesterday. However, you marked this PR as "ready for review" five days ago.
> 
> So is it ready for review or not? (Apparently it was not ready for review at least until yesterday?)
> 
> Is FTL-1670 now implemented completely, or will there be further commits?
> 
> This PR is now entirely different from what it was on Thursday (when it presumably was "ready for review"); will there be any further changes? If so, then maybe it is better to convert it back to draft.

I agree that MR should be mergeable before it's marked as "ready for review". Making many changes is making it harder for reviewers and asking for costly rereviews.

@inga-affinidi, let's be more tactful and considerate. Listen to the reason why it was done that way and let folks know how we can do things better.

sdk/core/src/CommonNetworkMember/BaseNetworkMember/RegisteringService.ts
inga-affinidi: > @anton-iskryzhytskyi doesn't have the same knowledge on the SDK that you have @inga-affinidi.

Different people have different knowledge, but they can share this knowledge during code review?

But also Anton _already_ refactored `generateFullSeed` and `parseDecryptedSeed` in this PR, so I don't see how this is about knowledge.

> Also, if you really consider that this refactor should be done right away, let's find a way to help @anton-iskryzhytskyi with them so he can focus on Sidetree.

What kind of help do you have in mind? The first comment in this thread already describes everything that could be done regarding this point.

If a code is needed, here is what I'm talking about (just one of ways of doing this, there are others):

* In `common-libs/common/src/shared/seedTools.ts`, replace the existing code:
  ```
  export const generateFullSeed = async (
    platformCryptographyTools: IPlatformCryptographyTools,
    didMethod: string,
    keyOptions?: KeyOptions,
    metadata?: Record<string, any>,
  ): Promise<string> => {
    const seed = await randomBytes(32)
    const seedHex = seed.toString('hex')

    // ... the remaining code
  }
  ```
  with
  ```
  export const generateFullSeed = async (
    platformCryptographyTools: IPlatformCryptographyTools,
    didMethod: string,
    keyOptions?: KeyOptions,
    metadata?: Record<string, any>,
  ): Promise<string> => {
    const seed = await randomBytes(32)
    const seedHex = seed.toString('hex')

    return generateFullSeedForSeedHex(platformCryptographyTools, seedHex, didMethod, keyOptions, metadata)
  }

  export const generateFullSeedForSeedHex = async (
    platformCryptographyTools: IPlatformCryptographyTools,
    seedHex: string,
    didMethod: string,
    keyOptions?: KeyOptions,
    metadata?: Record<string, any>,
  ): Promise<string> => {
    // ... the remaining code
  }
  ```

* In `RegisteringService.ts`, replace the existing code in `buildElemAnchoredSeed`
  ```
  const elemAnchoredSeedDraft = await generateFullSeed(
    this.platformCryptographyTools,
    this.didMethod,
    this.keyOptions,
    {
      anchoredDid: did,
    },
  )
  const originalSeed = originalFullSeed.split('++')[0]
  const draftSeed = elemAnchoredSeedDraft.split('++')[0]
  const elemAnchoredSeed = elemAnchoredSeedDraft.replace(draftSeed, originalSeed)
  ```
  with
  ```
  const { seedHex } = getDecryptedSeedParts(decryptedSeeds)
  const elemAnchoredSeed = await generateFullSeedForSeedHex(
    this.platformCryptographyTools,
    seedHex,
    this.didMethod,
    this.keyOptions,
    {
      anchoredDid: did,
    },
  )
  ```

> Let's not block him at this point, even if that means a new MR.

What is the purpose of this code review, then?
If the priority is to merge this anyway ASAP, without any additional changes, then we probably should just merge this PR as is, without reviewing, as soon as it stopped being a draft?

inga-affinidi: Although now that I look at this class (and related ones) more closely, I see that the current implementation is really hard to follow and that it can be much improved.

Current logic in this new file (it didn't exist prior to this PR) for all did methods except for `ELEM_ANCHORED_DID_METHOD`:
* Generate a new seed for this did method and `keyOptions`
  * It generates a new random `seedHex` and a bunch of random `externalKeys` according to `keyOptions`
* Encrypt it
* Create a `KeysService` instance for this encrypted seed
* Create a `DidDocumentService` instance for `KeysService` (it will decrypt the seed to extract did method and return a specific implementation for this did method)
* Create a did document by calling `buildDidDocument`
* Anchor did by calling `AnchoringService` for this encrypted seed and password and didDocument and nonce and did method (`AnchoringService` is also new and didn't exist prior to this PR)
  * nonce is always zero
  * encrypted seed and password are only used to obtain the decrypted seed
  * it only uses a first part (`seedHex`) of a decrypted seed (except that instead of just using `seedHex` it converts `seed` buffer to hex string, where `seed` buffer is obtained by converting `seedHex` from hex string to buffer in `parseDecryptedSeed`)
  * it always returns `void`
* Return did and encrypted seed

But the logic becomes even weirder when did method is `ELEM_ANCHORED_DID_METHOD`:
* Generate a new seed for `ELEM_DID_METHOD` and `keyOptions`
  * It generates a new random `seedHex` and a bunch of random `externalKeys` according to `keyOptions`
* Encrypt it
* Create a `KeysService` instance for this encrypted seed
* Create a `DidDocumentService` instance for `KeysService` (it will decrypt the seed to extract did method which is always `ELEM_DID_METHOD` and return a specific implementation for it)
  * `DidDocumentService` for `ELEM_DID_METHOD` uses `KeyVault`, and `KeyVault` decrypts seed and only uses `seedHex` and `externalKeys` (which were generated at random at the first step according to `keyOptions`)
* Create a did document by calling `buildDidDocument` (it seems that it adds public parts of `externalKeys` to the resulting document)
* Anchor did by calling `AnchoringService` for this encrypted seed and password and didDocument and nonce and `ELEM_DID_METHOD` (`AnchoringService` is also new and didn't exist prior to this PR)
  * nonce is always zero
  * for `ELEM_DID_METHOD` it does not use encrypted seed or password at all
  * it only uses a first part (`seedHex`) of a decrypted seed (except that instead of just using `seedHex` it converts `seed` buffer to hex string, where `seed` buffer is obtained by converting `seedHex` from hex string to buffer in `parseDecryptedSeed`)
  * it always returns `void`
* Take did and unencrypted seed
* Generate a _new_ seed for `ELEM_ANCHORED_DID_METHOD` and `keyOptions` and additional metadata (did from the previous step)
  * It generates a new random `seedHex` and a bunch of random `externalKeys` according to `keyOptions`
* Extract `seedHex` from the original seed and from the new seed
* Perform the string replacement (`.replace` method) on the new seed, replacing its `seedHex` with the original `seedHex`
* Discard the original seed (along with its `externalKeys`), discard the new `seedHex`
* Encrypt the new seed
* Return did and the encrypted seed.

This code is very complicated and hard to follow (it took me 40 minutes just to write a preceding part). The intent is almost impossible to deduce. And this is all in a new code.

It also looks like there is a bug: did document is created for the first seed, will contain public external keys from the first seed, but the second seed contains another set of external keys while the first set is discarded. So private keys corresponding to the public keys stored in the did document (which is a part of a returned seed, and which also was anchored) do not exist anymore; and public keys corresponding to the private external keys stored in returned seed (these private keys are used to sign data in some scenarios) are not actually made public.
Maybe this is not a bug but there are no comments to indicate that this is an intended behavior, and it does not look very intended.
And this is caused by the original code smell, when we generate one seed, do a bunch of things with it, and then generate another seed entirely (replacing its `seedHex` with the one from original seed) and discard the first one.

I can only attempt to guess what we were trying to achieve here: we needed seeds of this type to be created as the ordinary elem seeds, but also to contain a did in their metadata.
This could have been achieved in other easier to understand way:
* Treat ELEM_ANCHORED_DID_METHOD in the similar fashion as ELEM_DID_METHOD during the registration; get rid of "if it is ELEM_ANCHORED_DID_METHOD, use ELEM_DID_METHOD instead" conditions in code.
  * In `DidDocumentService`, make two separate methods for different purposes: `buildDidDocument` for registration, and `getDidDocument` for other purposes (post-registration)
* Instead of handling the entire encrypted seed during registration, only handle needed parts of unencrypted seed during registration;
* Only create one seed data with all the required data (except for metadata with did)
* After did is created, create a final seed (with did if needed)
  * These methods should be in `seedTools`. Instead of `generateSeed`, that should be e.g. `generateSeedData` (does not accept any metadata yet, returns an object) and `buildSeed` (accepts generated seed data plus metadata, returns a string)

anton-iskryzhytskyi:
All comments will be taken to attention and PR will be recreated as a draft and will be open for reviewers after consulting with Denys Popov

PR #145, opened 2021-09-15
[FTL-1836] Support for 'ecdsa' key type in seed generation

inga-affinidi:
This PR is equivalent to (a half of) https://gitlab.com/affinidi/foundational/cloud-wallet-api/-/merge_requests/120 , and should be in `affinidi-core-sdk` along with the keys and seed generation as a part of cloud-wallet-api upgrade to SDK v6.

### DIFF
--- a/common-libs/common/src/shared/seedTools.ts
+++ b/common-libs/common/src/shared/seedTools.ts
@@ -41,6 +41,23 @@ export const parseDecryptedSeed = (decryptedSeed: string) => {
   return { seed, didMethod, seedHexWithMethod, externalKeys, fullSeedHex }
 }
 
+const generateAdditionalKeys = async (cryptographyTools: IPlatformCryptographyTools, keyOptions: KeyOptions) => {
+  const filteredKeyTypes = keyOptions.keyTypes.flatMap((externalKeyType) => {
+    if (externalKeyType === 'ecdsa') {
+      return []
+    }
+
+    return [externalKeyType]
+  })
+
+  return Promise.all(
+    filteredKeyTypes.map(async (externalKeyType) => ({
+      externalKeyType,
+      keyInfo: await cryptographyTools.keyGenerators[externalKeyType](),
+    })),
+  )
+}
+
 export const generateFullSeed = async (
   platformCryptographyTools: IPlatformCryptographyTools,
   didMethod: string,
@@ -53,28 +70,19 @@ export const generateFullSeed = async (
     return seedWithMethod
   }
 
-  const additionalKeys = []
-  // eslint-disable-next-line no-restricted-syntax
-  for (const externalKeyType of keyOptions.keyTypes) {
-    if (externalKeyType === 'ecdsa') {
-      throw new Error('Please provide key type from the list: `rsa`, `bbs`. Some of your keys is not implemented!')
-    }
-
-    const { keyFormat, privateKey, publicKey } = await platformCryptographyTools.keyGenerators[externalKeyType]()
-
-    additionalKeys.push({
+  const additionalKeys = await generateAdditionalKeys(platformCryptographyTools, keyOptions)
+  const keysSeedSectionContent = additionalKeys.map(
+    ({ externalKeyType, keyInfo: { keyFormat, privateKey, publicKey } }) => ({
       type: externalKeyType,
       format: keyFormat,
       private: privateKey,
       public: publicKey,
       permissions: ['authentication', 'assertionMethod'],
-    })
-  }
+    }),
+  )
 
-  const keysSeedSection = base64url.encode(JSON.stringify(additionalKeys))
-
-  const fullSeed = `${seedWithMethod}++${keysSeedSection}`
-  return fullSeed
+  const keysSeedSection = base64url.encode(JSON.stringify(keysSeedSectionContent))
+  return `${seedWithMethod}++${keysSeedSection}`
 }
 
 export const convertDecryptedSeedBufferToString = (decryptedSeed: Buffer) => {
--- a/common-libs/common/tsconfig.json
+++ b/common-libs/common/tsconfig.json
@@ -3,7 +3,7 @@
     "module": "commonjs",
     "resolveJsonModule": true,
     "esModuleInterop": true,
-    "target": "ES2016",
+    "target": "ES2019",
     "noImplicitAny": true,
     "moduleResolution": "node",
     "sourceMap": true,
--- a/common-libs/common/CHANGELOG.md
+++ b/common-libs/common/CHANGELOG.md
@@ -1,3 +1,5 @@
+# 2.0.0-beta.9 (2021-09-15)
+* Support for ecdsa keys in seed generation
 # 2.0.0-beta.8 (2021-09-09)
 * DIDs are now resolved using `RegistryApiService` from `@affinidi/internal-api-clients`
 * Resolved DIDs are cached
--- a/common-libs/common/package-lock.json
+++ b/common-libs/common/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@affinidi/common",
-  "version": "2.0.0-beta.8",
+  "version": "2.0.0-beta.9",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "@affinidi/common",
-      "version": "2.0.0-beta.8",
+      "version": "2.0.0-beta.9",
       "license": "ISC",
       "dependencies": {
         "@affinidi/affinity-metrics-lib": "^0.0.25",
--- a/common-libs/common/package.json
+++ b/common-libs/common/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@affinidi/common",
-  "version": "2.0.0-beta.8",
+  "version": "2.0.0-beta.9",
   "description": "Common utilities and types for Affinidi SDKs",
   "main": "dist/index.js",
   "typings": "dist/index.d.ts",
--- a/sdk/browser/package-lock.json
+++ b/sdk/browser/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@affinidi/wallet-browser-sdk",
-  "version": "6.0.0-beta.11",
+  "version": "6.0.0-beta.12",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "@affinidi/wallet-browser-sdk",
-      "version": "6.0.0-beta.11",
+      "version": "6.0.0-beta.12",
       "license": "ISC",
       "dependencies": {
         "eccrypto-js": "^5.0.0",
--- a/sdk/browser/package.json
+++ b/sdk/browser/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@affinidi/wallet-browser-sdk",
-  "version": "6.0.0-beta.11",
+  "version": "6.0.0-beta.12",
   "description": "SDK monorepo for affinity DID solution for browser",
   "main": "dist/index.js",
   "typings": "dist/index.d.ts",
@@ -37,7 +37,7 @@
   ],
   "license": "ISC",
   "dependencies": {
-    "@affinidi/wallet-core-sdk": "^6.0.0-beta.11",
+    "@affinidi/wallet-core-sdk": "^6.0.0-beta.12",
     "eccrypto-js": "^5.0.0",
     "randombytes": "^2.1.0"
   },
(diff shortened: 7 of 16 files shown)


common-libs/common/src/shared/seedTools.ts
```diff
@@ -41,6 +41,23 @@ export const parseDecryptedSeed = (decryptedSeed: string) => {
...
 }
 
+const generateAdditionalKeys = async (cryptographyTools: IPlatformCryptographyTools, keyOptions: KeyOptions) => {
+  const filteredKeyTypes = keyOptions.keyTypes.flatMap((externalKeyType) => {
+    if (externalKeyType === 'ecdsa') {
+      return []
+    }
+
+    return [externalKeyType]
+  })
```
inga-affinidi, line 51: It is a more verbose and a bit less performant alternative to `keyOptions.keyTypes.filter((externalKeyType) => externalKeyType !== 'ecdsa')`.

Unfortunately, `filter` in TS does not know about predicates and type narrowing, so the result of a call to `filter` would still be typed as an array of all possible key types, while the result of a call to `flatMap` is typed as an array of all key type except 'ecdsa'.

common-libs/common/tsconfig.json
```diff
@@ -3,7 +3,7 @@
     "module": "commonjs",
     "resolveJsonModule": true,
     "esModuleInterop": true,
-    "target": "ES2016",
+    "target": "ES2019",
```
inga-affinidi, line 6: (for `Array.prototype.flatMap` support)

vitalyrudenko-affinidi (approved):

common-libs/common/tsconfig.json
vitalyrudenko-affinidi: Does this mean that the final build of the SDK won't be supported in browsers that don't support ES2019? (not that we really care about that, as I see).

common-libs/common/src/shared/seedTools.ts
vitalyrudenko-affinidi: Can't we assert type here? With `Omit` or something. Or you want it to be implicit?

common-libs/common/tsconfig.json
inga-affinidi: Some other packages already rely on ES2019.

If we care about old browsers, we should probably use some polyfills (and `flatMap` is trivially implemented using `map` and `flatten`) instead of trying to limit ourselves to outdated APIs.

common-libs/common/src/shared/seedTools.ts
inga-affinidi: `externalKeyType as Exclude<KeyType, 'ecdsa'>` is not type safe, just like any other assertion; it would be basically us telling TS that we know better than the compiler, and that we trust our own code and swear `externalKeyType` cannot be equal to 'ecdsa' here.

I'd prefer to not have to swear and not to trust my code.

inga-affinidi:
@anton-iskryzhytskyi , of course, forgot that you were working on this file too.

common-libs/common/src/shared/seedTools.ts
anton-iskryzhytskyi: @vitalyrudenko-affinidi @inga-affinidi good example of fighting with TS.
@inga-affinidi I prefer to write a test instead of this

inga-affinidi: @anton-iskryzhytskyi , it is not fighting TS, it is leveraging TS abilities.

dmfilipenko-affinidi: I guess something like this can be used. But, as @inga-affinidi mention it's not type safe
```typescript
const filteredKeyTypes = keyOptions.keyTypes.filter<'rsa' | 'bbs'>((externalKeyType): externalKeyType is ('rsa' | 'bbs') => externalKeyType !== 'ecdsa')
```

anton-iskryzhytskyi (commented):
Let's consider writing tests instead of relying on TS. What do you think folks

common-libs/common/src/shared/seedTools.ts
inga-affinidi: @dmfilipenko-affinidi , if this was a repeating pattern we could write something like
```
  const isNot = <TExcluded extends string | number | boolean>(excludedValue: TExcluded) => {
    return <TValue extends string | number | boolean>(value: TValue | TExcluded): value is Exclude<TValue, TExcluded> => value !== excludedValue
  }
  const filteredKeyTypes = keyOptions.keyTypes.filter(isNot('ecdsa'))
```

(Also there are `without` methods in lodash and in underscore.js but for some reason they are not typed properly)

anton-iskryzhytskyi: What about writing tests for that instead of using some good typed functions? Even though it will be good-typed it can allow error, in the same time tests give us more durability

vitalyrudenko-affinidi: I do think that in some other parts of SDK types have become too complex and cumbersome to maintain, but some of those things are exposed to the public, so we need good typings for that.
And tests are good even when typescript covers some of these cases (because I don't think that types are very reliable - someone can just write `as any` here and we will lose any checks in this code).

Here I'm not sure how this tool is used, but I suppose it's an internal function. Maybe it is better to write a small test and assert types here. If I come here in a year, I might not remember why we use `flatMap` here instead of `filter` and remove it - and who knows how many bugs we'll encounter because of this minor change. Tests will ensure that developers know WHY this is important.

inga-affinidi: _Both_ test and types are important.

We're using TS for a reason; it allows us to catch obvious mistakes immediately at the time of writing, while with unit tests the feedback is delayed (and with integration tests it is delayed even more).

That to be said, I'm not sure if it makes sense to write unit tests for every single line, that often degenerates into writing unit tests for the sake of writing unit tests, where we write 20 lines of test code in order to test that one line of actual code indeed does what it says, and then update hundreds of lines of test code whenever an actual code is slightly refactored without any functional changes (I've seen the examples).

> because I don't think that types are very reliable - someone can just write as any here and we will lose any checks in this code

Writing `as any` is basically the same as saying "I won't write any tests for this code because I know better" and adding coverage ignore comments, but for types. So by that logic tests are also not very reliable because someone can just not write them.

But also explicit `any` usage should be an eslint error (unfortunately it is not in the old version of eslint-config SDK uses for now), and every attempt to disable it should be explicitly justified in code.

> If I come here in a year, I might not remember why we use flatMap here instead of filter and remove it - and who knows how many bugs we'll encounter because of this minor change

If you just replace `flatMap` with `filter` here, your code won't compile because you won't be able to access `cryptographyTools.keyGenerators[externalKeyType]` below (because 'ecdsa' cannot be used to index `keyGenerators`).
If you then fix it with a type assertion so that the code will compile, it will work, and this change by itself will not introduce any new bugs; but it will make it easier to introduce new bugs in the future.

But then we have eslint rule which disincentivizes using type assertions, so presumably that'll make you suspect that `flatMap` was there for a reason (because when getting rid of `flatMap` you had to write a type assertion _and_ get a new ESLint violation).

> Tests will ensure that developers know WHY this is important.

Why what exactly is important?

vitalyrudenko-affinidi: Good points.

> Why what exactly is important?

I mean this whole "excluding ecdsa key type" thing.

vitalyrudenko-affinidi: > where we write 20 lines of test code in order to test that one line of actual code indeed does what it says

Don't you think that we have a lot of types that cover just few lines of code? This argument goes both ways.

On the other hand, I don't like how typescript works just because JS itself is very flexible, therefore covering it with types is hard. I now need to be an **expert** in TS just to understand what happens in the SDK.

vitalyrudenko-affinidi: Well, just to clarify, I don't say that we should abandon TS, but I do say that we should be more forgiving when TS is very strict and instead write a couple of tests + perform runtime validation for inputs if we don't trust them.

vitalyrudenko-affinidi: > In order to understand what happens inside SDK, in order to develop it, you need to be experienced with the language SDK is written in. Which happens to be TS.

Yeah why do our developers need to be experts in TypeScript to develop the SDK? Isn't it kind of arbitrary?
We're just limiting our developers, because they need to learn TS for something that won't even make it to the final build of SDK, just for compile-time checks, which can be covered by tests - and those are quite easily understandable by everyone - not just TS experts.

I don't have examples of huge types for the sake of types (there are a lot of them in SDK), but I do have this screenshot I made recently that blows my mind:
photo_2021-09-09_18-14-24

I thought everyone's okay with these complex types, but now I see that there are more people that are confused by this, so maybe we should reconsider this decision (of very strict typing) - not here and not right now, of course.

vitalyrudenko-affinidi: E.g. I haven't had much experience with TS previously, so now it's very hard for me to contribute to the SDK, and it becomes harder every day - I'm having hard time even reviewing latest PRs. I don't say that we should do something just because of me, but I do want to point out that "knowing TS" is not a very important skill for a good developer.

vitalyrudenko-affinidi: Anyway, it's just a rant, I'm okay with current code in this PR, that's why I approved it. But I'm willing to discuss any concerns that my colleagues have.

inga-affinidi: > I don't have examples of huge types for the sake of types (there are a lot of them in SDK), but I do have this screenshot I made recently that blows my mind:

Of course this code has no place in "userspace" part of SDK. It is more of a generic utility code (it could even exist in a separate package, not related in any way to the rest of affinidi code).
And it is here precisely to make writing userspace code easier, to improve developer experience for developers who are developing business features and not implementing generic openapi support. So that instead of
```
const headers: any = {
  Authorization: `Bearer ${token}`,
  ...(this._storageRegion ? { ['X-DST-REGION']: this._storageRegion } : {}),
}

const { skip, limit } = paginationOptions

const url = `${this._vaultUrl}/data/${skip}/${skip + limit - 1}`

const { body: blobs } = await this._api.execute(null, {
  url,
  headers,
  method: 'GET',
})

return blobs
```
(code from SDK v5) without any autocompletion, type checking etc, with any mistype only uncovered by integration tests or in production, we could do
```
return this.client.GetCredentials({
  authorization: `Bearer ${accessToken}`,
  storageRegion,
  pathParams: { start: `${start}`, end: `${end}` },
})
```
where every single word supports auto completion and type checking.

Doesn't that change improve developers' experience? As long as they don't peek under the hood and try to make sense of open api parser, which, being entirely in TS by its nature (in order to make the experience above possible), indeed requires TS knowledge.

vitalyrudenko-affinidi: > Could that be because latest PRs for the most part consisted of changes to this generic code? It shouldn't change often; it is quite likely that it won't change at all anymore, so you won't have to review code like that in the future.

Well, if that's the case, then I'm okay with having complex types in few places.  
But we shouldn't decrease the quality of our code to "leverage TS advantages", otherwise we'll have a fragile code that is hard to maintain and requires TS knowledge to understand its purpose.

> Knowing JS is also not a very important skill for a good developer, but I think it's a reasonable expectation for a developer who intends to contribute to SDK? :)

I suppose SDK is in JS because it's most popular language and it can be easily used in browser, react-native, node, etc. Not because it's a good skill, it's a requirement. That I see as a good reason to use JS.
Reason to use TS, on the other hand, is a bit blurry for me. A lot of types just for autocomplete and compile-time checks (which can and should be covered by tests anyway).
I do agree that TS is very useful, but to an extent. And we should have some limits on how deep/strict we want to go.

inga-affinidi: Maybe it will be easier if you look at these three ways of saying the same thing (JS-wise):

```
const filteredKeyTypes = keyTypes.filter((keyType) => keyType !== 'ecdsa')
const filteredKeyTypes = keyTypes.filter((keyType) => keyType !== 'ecdsa' ? true : false)
const filteredKeyTypes = keyTypes.flatMap((keyType) => keyType !== 'ecdsa' ? [keyType] : [])
```

The first one is the most concise and the most direct, but I wouldn't say the other two are fragile or harder to maintain or require TS knowledge.

(TS knowledge is only needed to know that only the third one [narrows](https://www.typescriptlang.org/docs/handbook/2/narrowing.html) the elements of the resulting array by excluding 'ecdsa' type without having to explicitly write any type predicates or assertions)

DenisPopov15 (approved):

PR #148, opened 2021-09-21
FTL-1670 anchor and store did to seed metadata

anton-iskryzhytskyi:
Previous PR was closed due to unclear commit history

This is a part of FTL-1670, adding the future to keep creation of dids (method elem) on the backend through registry-service
This PR provides changes in common lib:
*  to work with updated format of seed with metadata (including fix of old format with external keys (base64 can include ++)
* to work with new internal seed method - elem-anchored (part that responsible for creating such seeds will be provided in next PR)
* improving codebase related to DidDocumentService

- [x] - describe PR what, why, how
- [x] - fix remarks from previous PR


common-libs/common/src/services/DidDocumentService/index.ts
```diff
@@ -9,40 +10,24 @@ export { KeyVault } from './KeyVault'
...
-      did = this.getMyDid()
-    }
-
-    const signingKey = 'primary'
-
-    return `${did}#${signingKey}`
-  }
+  static createDidDocumentService(
+    keysService: KeysService,
+  ): JoloDidDocument | ElemDidDocument | ElemAnchoredDidDocument {
```
inga-affinidi, line 22: By the way TS should be able to automatically infer the type there.
Or you can use `ConcreteDidDocument`.

common-libs/common/src/services/DidDocumentService/ConcreteDidDocument.ts
```diff
@@ -0,0 +1,7 @@
+import { DidResolver } from '../../shared/DidResolver'
+
+export interface ConcreteDidDocument<T extends { id: string } = { id: string }> {
```
inga-affinidi, line 3: Not sure why do we need this generic parameter here. It would probably be enough to declare the return value of `buildDidDocument` as `Record<string, unknown> & { id: string }`.

Also I'm not sure about the naming.
This is not a did document; this is a did document factory? wrapper? something else? Did document is something that `buildDidDocument` returns, but not the object with that method.
And this is an interface, as far from "concrete" as it could be.

common-libs/common/test/factory/didFactory.ts
```diff
@@ -9,7 +9,7 @@ interface TestDid {
   encryptedSeed: string
   seedHex: string
   did: string
-  didDocument: { id: string }
+  didDocument: { id: string } | any | any
```
inga-affinidi, line 12: One would be enough here

sdk/core/src/services/AnchoringService.ts
```diff
@@ -0,0 +1,73 @@
+import { ELEM_ANCHORED_DID_METHOD, ELEM_DID_METHOD, JOLO_DID_METHOD } from '../_defaultConfig'
+import { RegistryApiService } from '@affinidi/internal-api-clients'
+import { KeysService } from '@affinidi/common'
+
+export class AnchoringService {
```
inga-affinidi, line 5: Not sure if this even should be called a service.
The only way it's used is: create an instance, and immediately, in the next line, call `anchorDid`.
Might make sense to just `export const anchorDid(api, encryptedSeed, ...)` here.

sdk/core/src/services/RegisteringService.ts
```diff
@@ -0,0 +1,91 @@
...
+  generateFullSeed,
+  KeysService,
+  joinSeedWithMethodAndBase64EncodedData,
+  generateSeedHexWithMethod,
+} from '@affinidi/common'
+import { AnchoringService } from './AnchoringService'
+import { RegistryApiService } from '@affinidi/internal-api-clients'
+import { buildBase64EncodedAdditionalData } from '@affinidi/common/dist/shared/seedTools'
+
+export class RegisteringService {
```
inga-affinidi, line 15: Same as with AnchoringService

inga-affinidi:
I'd say that everything but the last commit ("feat: FTL-1670 add support of elem-anchored method") is ready to be merged, in principle.

Maybe we should split this into two PRs (this PR with all commits except for the last one plus moving anchoring / registering into separate modules _without changing their actual code_; and another PR with the actual FTL-1670 implementation) to merge the first one now so that review of these changes won't block the second one and so that the second one is easier to review.

common-libs/common/src/shared/seedTools.ts
```diff
@@ -10,35 +10,52 @@ export type KeyOptions = {
...
-  const { seedHex, didMethod, base64EncodedKeys } = getDecryptedSeedParts(decryptedSeed)
+export const parseDecryptedSeed = (decryptedSeed: string): ParseDecryptedSeedResult => {
+  const { seedHex, didMethod, base64EncodedKeys, base64EncodedAdditionalData } = getDecryptedSeedParts(decryptedSeed)
 
   validateDidMethodSupported(didMethod)
   const seedHexWithMethod = `${seedHex}++${didMethod}`
   const seed = Buffer.from(seedHex, 'hex')
+  const additionalData = base64EncodedAdditionalData && JSON.parse(base64url.decode(base64EncodedAdditionalData))
+  const parsedFromBase64ExternalKeys = base64EncodedKeys && JSON.parse(base64url.decode(base64EncodedKeys))
+  const externalKeys = additionalData?.[EXTERNAL_KEYS_KEY] || parsedFromBase64ExternalKeys
```
inga-affinidi, line 55: `??` is more appropriate than `||` here.

common-libs/common/src/services/DidDocumentService/ElemAnchoredDidDocument.ts
```diff
@@ -0,0 +1,28 @@
...
+
+  getMyDid(): string {
+    const { anchoredDid } = this._keyVault.metadata
+    return `${anchoredDid}`
+  }
+  getKeyId(did: string = null) {
+    return `${did ? did : this.getMyDid()}#${this._signingKey}`
+  }
+
+  async buildDidDocument(didResolver?: DidResolver) {
```
inga-affinidi, line 21: Can we make this parameter non-optional in this PR?

anton-iskryzhytskyi: I'm not sure how to do it elegantly, as in case we make it non-optional it means it is required for each buildDidDocument even if it is not related to ElemAnchoredDidDocument
I don't have a good solution for it and finding out will take a time

inga-affinidi: Yes, that would mean it is required for every call to `buildDidDocument`, because the caller cannot and should not know what kind of `DidDocumentService` it uses. That would be much more elegant than forcing callers to know when it is safe and when it is unsafe to call `buildDidDocument` without arguments.

It seems that, aside from possible tests, `buildDidDocument` is only used in an instance method of `Affinity` (so it can just pass `this._didResolver`) and in `BaseNetworkMember.register` (where we have all the data needed to create a `DidResolver` instance).

inga-affinidi:
And one more thing: we'll need to bump the package versions here

YuriiSkrypnyk (dismissed):
LGTM, just one comment

common-libs/common/CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+# 2.0.0-beta.1 (2021-09-22)
```
inga-affinidi, line 1: -beta.10

sdk/core/package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@affinidi/wallet-core-sdk",
-  "version": "6.0.0-beta.13",
+  "version": "6.0.0-beta.14",
```
inga-affinidi, line 3: and also on line 9

common-libs/common/package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@affinidi/common",
-  "version": "2.0.0-beta.9",
+  "version": "2.0.0-beta.10",
```
inga-affinidi, line 3: and also on line 9

inga-affinidi (approved):

YuriiSkrypnyk (approved):

PR #149, opened 2021-09-22
LocalExpiringDidAuthResponseToken for token expiration calculation

ezgi-affinidi:
Used LocalExpiringDidAuthResponseToken class from did-auth for token expiration calculation

### DIFF
--- a/common-libs/internal-api-clients/src/helpers/DidAuthAdapter.ts
+++ b/common-libs/internal-api-clients/src/helpers/DidAuthAdapter.ts
@@ -20,5 +20,4 @@ export class DidAuthAdapter {
     const localExpiringDidAuthResponseTokenCheck = LocalExpiringDidAuthResponseToken.initialize(tokenRequestTime, token)
     return localExpiringDidAuthResponseTokenCheck.isExpiredAt(Date.now())
   }
-
 }
--- a/common-libs/internal-api-clients/src/helpers/DidAuthAdapter.ts
+++ b/common-libs/internal-api-clients/src/helpers/DidAuthAdapter.ts
@@ -1,4 +1,4 @@
-import { DidAuthService } from '@affinidi/affinidi-did-auth-lib'
+import { DidAuthService, LocalExpiringDidAuthResponseToken } from '@affinidi/affinidi-did-auth-lib'
 import { AffinidiDidAuthServiceOptions } from '@affinidi/affinidi-did-auth-lib/dist/DidAuthService/DidAuthService'
 
 export class DidAuthAdapter {
@@ -16,7 +16,8 @@ export class DidAuthAdapter {
     return this._didAuthService.createDidAuthResponseToken(didAuthRequestToken)
   }
 
-  isTokenExpired(token: string): boolean {
-    return this._didAuthService.isTokenExpired(token)
+  isTokenExpired(token: string, tokenRequestTime: number): boolean {
+    const localExpiringDidAuthResponseTokenCheck = LocalExpiringDidAuthResponseToken.initialize(tokenRequestTime, token)
+    return localExpiringDidAuthResponseTokenCheck.isExpiredAt(Date.now())
   }
 }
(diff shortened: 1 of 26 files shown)


inga-affinidi:
It doesn't build; you probably need to update `did-auth-lib` dependency in `package.json` files and re-run `npm run bootstrap` in the root project directory (and then replace tabs (caused by lerna bug) with spaces in regenerated package-lock.json)

common-libs/internal-api-clients/src/helpers/DidAuthAdapter.ts
```diff
@@ -17,6 +18,7 @@ export class DidAuthAdapter {
   }
 
   isTokenExpired(token: string): boolean {
-    return this._didAuthService.isTokenExpired(token)
+    const LocalExpiringDidAuthResponseTokenCheck = LocalExpiringDidAuthResponseToken.initialize(this.tokenRequestTime, token)
```
inga-affinidi, line 21: We only use names starting with capital letters for classes and types (and constants which are SNAKE_CASE).

`npm run lint` should have complained about this.

Just rename it to `const localExpiringDidAuthResponseTokenCheck`

common-libs/internal-api-clients/src/helpers/DidAuthAdapter.ts
```diff
@@ -1,8 +1,9 @@
-import { DidAuthService } from '@affinidi/affinidi-did-auth-lib'
+import { DidAuthService, LocalExpiringDidAuthResponseToken } from '@affinidi/affinidi-did-auth-lib'
 import { AffinidiDidAuthServiceOptions } from '@affinidi/affinidi-did-auth-lib/dist/DidAuthService/DidAuthService'
 
 export class DidAuthAdapter {
   private readonly _didAuthService: DidAuthService
+  public tokenRequestTime!: number
```
inga-affinidi, line 6: I _really_ don't like the idea of storing some state in this service. Especially one that is modifiable by external code (public non-readonly field).

`DidAuthSession` seems to be a more proper place for that.

inga-affinidi: (because it corresponds to an actual session, and it stores the response token, so it makes sense to store a request time there as well)

common-libs/internal-api-clients/src/helpers/DidAuthManager.ts
```diff
@@ -2,10 +2,13 @@ import { DidAuthAdapter } from '../helpers/DidAuthAdapter'
 
 export const createDidAuthSession = (didAuthAdapter: DidAuthAdapter) => {
   let responseToken: string | undefined
+  let tokenRequestTime!: number
 
   return {
     async getResponseToken(createRequestToken: (did: string) => Promise<string>) {
-      if (!responseToken || didAuthAdapter.isTokenExpired(responseToken)) {
+      if (!responseToken || didAuthAdapter.isTokenExpired(responseToken, tokenRequestTime)) {
+        tokenRequestTime = Date.now()
```
inga-affinidi, line 10: Or alternatively:

* Store `let responseTokenInfo: { tokenRequestTime: string; responseToken: string } | undefined`
* To determine whether you should request new token: check that responseTokenInfo is non-empty or that token is not expired
* `const tokenRequestTime = Date.now()`
* `const responseToken = await didAuthAdapter.createDidAuthResponseToken(...)`
* and in the end, `responseTokenInfo = { tokenRequestTime, responseToken }`

This way, you'll only have one nullable variable; and both you and TS will know that if it's not empty, then both token and request time are non-empty.
Right now, from TS standpoint (and other developers', if they're not reading the implementation really carefully) it is possible that `responseToken` is non-empty yet `tokenRequestTime` is empty (which is probably why you had to declare it with `!` which is a way of saying that you guarantee it will never be empty; but this is wrong: it will be empty prior to the first request to `getResponseToken`)

sdk/core/CHANGELOG.md
```diff
@@ -1,3 +1,5 @@
+# release 6.0.0-beta.14 (2021-09-29)
```
inga-affinidi, line 1: But your version is "beta.15"...

inga-affinidi (approved):

anton-iskryzhytskyi (approved):

PR #150, opened 2021-09-22
FTL-1670 update base network member to anchor and store seed metadata

anton-iskryzhytskyi:
This PR extending [previous](https://github.com/affinityproject/affinidi-core-sdk/pull/148) and adds the main feature of FTL-1670 - creating did on the backend


sdk/core/src/CommonNetworkMember/BaseNetworkMember.ts
```diff
@@ -209,20 +206,21 @@ export abstract class BaseNetworkMember {
...
+    const registeringService = new RegisteringService(
+      api,
+      didMethod,
+      dependencies.platformCryptographyTools,
+      password,
+      keyOptions,
+    )
 
-    return { did, didDocumentKeyId, encryptedSeed }
+    return registeringService.register()
```
inga-affinidi: Again, there is no point in calling this a "service", just a method `export const register` would suffice.

Even if we want to call this a "service", it should be reusable. What point is there to have a "service" that only has one instance method that can only be called once?
For example, its constructor could accept `api` and `platformCryptographyTools`, while `register` method would accept `didMethod`, `password` and `keyOptions` (parameters specific to this registration)

sdk/core/src/CommonNetworkMember/BaseNetworkMember.ts
```diff
@@ -233,43 +231,8 @@ export abstract class BaseNetworkMember {
...
-    // NOTE: for metrics purpose in case of ELEM method
-    if (didMethod === ELEM_DID_METHOD) {
-      try {
-        await api.anchorDid({ did, didDocumentAddress: '', ethereumPublicKeyHex: '', transactionSignatureJson: '' })
-      } catch (error) {
-        console.log('to check logs at the backend', error)
-      }
-    }
+    const anchoringService = new AnchoringService(api, encryptedSeed, password, didDocument, nonce)
+    return anchoringService.anchorDid()
```
inga-affinidi: See comments for `RegisteringService` above

sdk/core/src/services/AnchoringService.ts
```diff
@@ -0,0 +1,73 @@
...
+      didDocumentAddress,
+      ethereumPublicKeyHex,
+      transactionSignatureJson,
+      nonce: this.nonce,
+    })
+
+    return { did: response?.body?.did }
+  }
+
+  private async anchorDidWithElemMethod(anchoredDidElem: boolean) {
```
inga-affinidi: It would be great if you could split this into two separate commits. One with extracting anchoring into a separate file, and another with changing its logic.

Right now it is really hard to see how flow has actually changing, because this commit is basically "remove a bunch of lines from one file, add a bunch of other lines to another file".

sdk/core/src/services/RegisteringService.ts
```diff
@@ -0,0 +1,91 @@
...
+
+  private get passwordBuffer() {
+    return KeysService.normalizePassword(this.password)
+  }
+
+  private createAnchoringService(encryptedSeed: string, didDocument: any) {
+    return new AnchoringService(this.api, encryptedSeed, this.password, didDocument, 0)
+  }
+
+  private async registerJoloOrElem() {
```
inga-affinidi: Same as for `AnchoringService`

sdk/core/src/services/anchoringHandler.ts
```diff
@@ -0,0 +1,65 @@
...
+      transactionSignatureJson = await keysService.createTransactionSignature(digestHex, seedHex)
+    }
+
+    const transactionPublicKey = KeysService.getAnchorTransactionPublicKey(seedHex, didMethod)
+    const ethereumPublicKeyHex = transactionPublicKey.toString('hex')
+
+    await api.anchorDid({ did, didDocumentAddress, ethereumPublicKeyHex, transactionSignatureJson, nonce })
+  }
+
+  if (didMethod === ELEM_DID_METHOD && anchoredDidElem) {
```
DenisPopov15, line 40: @anton-iskryzhytskyi What is the different between this and next method only if it return or not return something, correct? Why not return always, and on level above figure out if that value which returned need to be used or not (and that if there present anyway I believe)

sdk/core/src/services/anchoringHandler.ts
```diff
@@ -0,0 +1,65 @@
...
+import { KeysService } from '@affinidi/common'
+
+export const anchorDid = async (
+  api: RegistryApiService,
+  encryptedSeed: string,
+  password: string,
+  didDocument: any,
+  anchoredDidElem: boolean = false,
+  nonce: number = 0,
+): Promise<{ did: string } | undefined> => {
```
YuriiSkrypnyk, line 12: This function looks inconsistent to me.
1. Is it expected that we return only in case  `ELEM_DID_METHOD` and `anchoredDidElem`? We could return did in any case and then return type will be `Promise<{ did: string }>`.
2. We could initialise `anchorDidParams` and pass then where it is needed to avoid duplication:
```
const anchorDidParams = {
      did,
      didDocumentAddress: '',
      ethereumPublicKeyHex: '',
      transactionSignatureJson: '',
      anchoredDidElem,
 }
...
await api.anchorDid(anchorDidParams)
...
await api.anchorDid(anchorDidParams)
```
3. Why 3rd call to `api.anchorDid` wrapped by try-catch and first 2 not?

dmfilipenko-affinidi, line 12: Why not return `null`, rather than `undefined`?

sdk/core/src/services/anchoringHandler.ts
```diff
@@ -0,0 +1,58 @@
...
+}
+
+export const anchorDid = async (
+  api: RegistryApiService,
+  encryptedSeed: string,
+  password: string,
+  didDocument: any,
+  anchoredDidElem: boolean = false,
+  nonce: number = 0,
+): Promise<{ did: string } | undefined> => {
```
dmfilipenko-affinidi, line 50: Are you sure this type is correct? Shouldn't be something like this `Promise<{did: string | undefined}>`. In this case, I would prefer to use null rather than undefined, but if you are using undefined across the code base, better to keep it for consistency.

sdk/core/src/CommonNetworkMember/BaseNetworkMember.ts
```diff
@@ -210,26 +206,14 @@ export abstract class BaseNetworkMember {
...
-        sdkVersion: extractSDKVersion(),
-      }),
-    )
-    const did = didDocument.id
-    const didDocumentKeyId = didDocumentService.getKeyId()
-
-    await BaseNetworkMember._anchorDid(encryptedSeed, password, didDocument, 0, options)
-
-    return { did, didDocumentKeyId, encryptedSeed }
+    return register(api, didResolver, didMethod, dependencies.platformCryptographyTools, password, keyOptions)
```
dmfilipenko-affinidi, line 216: Are we handling error which thrown by this method or it consumer responsibility?

sdk/core/src/services/anchoringHandler.ts
anton-iskryzhytskyi: here should be { did: string } only, seems I missed that

sdk/core/src/CommonNetworkMember/BaseNetworkMember.ts
anton-iskryzhytskyi: yes, as it was before. The main goal of this PR to add an 'elem-anchored' feature, not make refactoring

sdk/core/src/services/anchoringHandler.ts
```diff
@@ -0,0 +1,58 @@
...
+export const anchorDid = async (
+  api: RegistryApiService,
+  encryptedSeed: string,
+  password: string,
+  didDocument: any,
+  anchoredDidElem: boolean = false,
+  nonce: number = 0,
+): Promise<{ did: string }> => {
+  const response = await api.anchorDid({
+    ...(await getAnchoringParams(api, encryptedSeed, password, didDocument, nonce)),
```
dmfilipenko-affinidi, line 52: Potentially can fail if `getAnchoringParams` method will respond with something that is not an object or array.

dmfilipenko-affinidi (approved):

sdk/core/src/services/anchoringHandler.ts
YuriiSkrypnyk: yes, `getAnchoringParams()` still have chance to return undefined if `didMethod` is not `JOLO_DID_METHOD` or `ELEM_DID_METHOD`. Better throw some expected error like `This did method is not supported` or return empty object than have runtime error here

inga-affinidi (changes requested):

sdk/core/src/services/anchoringHandler.ts
anton-iskryzhytskyi: in case of wrong did method this function returns Promise<undefined>, resolved value is undefined, but distraction operator has identical behavior for undefined, null, {}, 0 etc.

But anyway now I prefer to add throwing error here.

sdk/core/src/services/registeringHandler.ts
```diff
@@ -0,0 +1,94 @@
...
+  didResolver: DidResolver,
+  password: string,
+  platformCryptographyTools: IPlatformCryptographyTools,
+  keyOptions?: KeyOptions,
+) => {
+  const seedWithHexMethod = await generateSeedHexWithMethod(ELEM_DID_METHOD)
+  const { encryptedSeed, keysService } = await KeysService.fromSeedAndPassword(seedWithHexMethod, password)
+
+  const didElemDocumentService = DidDocumentService.createDidDocumentService(keysService)
+  const didDocument = await didElemDocumentService.buildDidDocument(didResolver)
```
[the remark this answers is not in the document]
anton-iskryzhytskyi: good point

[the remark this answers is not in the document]
anton-iskryzhytskyi: now did:elem is generated with keys, then after anchoring of did it is added to additional data, that is all

sdk/core/src/services/registeringHandler.ts
```diff
@@ -0,0 +1,94 @@
...
+  const { did } = await anchorDid(api, encryptedSeed, password, didDocument, true)
+
+  const elemAnchoredSeed = await buildElemAnchoredSeed(did, seedWithHexMethod, platformCryptographyTools, keyOptions)
+
+  const {
+    keysService: elemAnchoredKeysService,
+    encryptedSeed: elemAnchoredEncryptedSeed,
+  } = await KeysService.fromSeedAndPassword(elemAnchoredSeed, password)
+  const didDocumentService = DidDocumentService.createDidDocumentService(elemAnchoredKeysService)
+  const didDocumentKeyId = didDocumentService.getKeyId()
```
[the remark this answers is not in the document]
anton-iskryzhytskyi: no, as getKeyId of ElemAnchored returns anchored value

sdk/core/src/services/registeringHandler.ts
```diff
@@ -0,0 +1,94 @@
...
+  api: RegistryApiService,
+  didResolver: DidResolver,
+  password: string,
+  platformCryptographyTools: IPlatformCryptographyTools,
+  keyOptions?: KeyOptions,
+) => {
+  const seedWithHexMethod = await generateSeedHexWithMethod(ELEM_DID_METHOD)
+  const { encryptedSeed, keysService } = await KeysService.fromSeedAndPassword(seedWithHexMethod, password)
+
+  const didElemDocumentService = DidDocumentService.createDidDocumentService(keysService)
```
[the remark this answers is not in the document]
anton-iskryzhytskyi: https://github.com/affinityproject/affinidi-core-sdk/pull/150/files#diff-a2a35df7a7951c31177c2c72d4c68da76cb5edefe6f628026afd95309896e239R48 
Here is anchoring of did in eth ropsten network with sidetree on affinity-registry side. I updated the name of variable to make it easier for people to understand why this did is added to seed as a metadata.anchoredDid.

sdk/core/src/services/registeringHandler.ts
```diff
@@ -0,0 +1,94 @@
...
+  platformCryptographyTools: IPlatformCryptographyTools,
+  password: string,
+  keyOptions?: KeyOptions,
+) => {
+  if ([ELEM_DID_METHOD, JOLO_DID_METHOD].includes(didMethod)) {
+    return registerJoloOrElem(api, didResolver, didMethod, password, platformCryptographyTools, keyOptions)
+  }
+
+  if (ELEM_ANCHORED_DID_METHOD === didMethod) {
+    return registerElemAnchored(api, didResolver, password, platformCryptographyTools, keyOptions)
```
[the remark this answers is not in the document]
anton-iskryzhytskyi: I don't quite understand your explanation, could you please add a commit for that, if you really think it should be included and this update could stop this PR from merging

sdk/core/test/integration/NetworkMember.test.ts
```diff
@@ -239,6 +241,24 @@ describe('CommonNetworkMember', () => {
     expect(didMethod).to.be.equal(elemDidMethod)
   })
 
+  it('.register (elem-anchored did method)', async () => {
```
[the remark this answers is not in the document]
anton-iskryzhytskyi: yes, make sense.

sdk/core/src/services/anchoringHandler.ts
```diff
@@ -0,0 +1,60 @@
+import { ELEM_DID_METHOD, JOLO_DID_METHOD } from '../_defaultConfig'
+import { RegistryApiService } from '@affinidi/internal-api-clients'
+import { DidDocumentService, KeysService } from '@affinidi/common'
+
+const getAnchoringParams = async (
```
inga-affinidi, line 5: The name of this method is misleading; it does not just return anchoring params, but it also stores didDocument in IPFS.

sdk/core/src/services/registeringHandler.ts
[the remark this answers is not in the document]
inga-affinidi: This (somewhat) answers one part of my question:

> It is unclear where would registry service obtain the document

It does not answer another:

> In any way, it seems that registry service does not generate DID documents; SDK still does that. And the primary stated objective of FTL-1670 is to move document **generation** out of SDK to the registry service.

sdk/core/src/services/registeringHandler.ts
```diff
@@ -0,0 +1,79 @@
...
+) => {
+  const fullElemSeed = await generateFullSeed(platformCryptographyTools, ELEM_DID_METHOD, keyOptions)
+  const { encryptedSeed, keysService } = await KeysService.fromSeedAndPassword(fullElemSeed, password)
+
+  const didElemDocumentService = DidDocumentService.createDidDocumentService(keysService)
+  const didDocument = await didElemDocumentService.buildDidDocument(didResolver)
+
+  const { did: anchoredInBlockchainDid } = await anchorDid(api, encryptedSeed, password, didDocument, true)
+
+  const elemAnchoredSeed = processAnchoredElemDidSeed(keysService.decryptSeed(), anchoredInBlockchainDid)
```
inga-affinidi, line 50: No need to `decryptSeed` here, you already have `fullElemSeed`

common-libs/common/src/shared/seedTools.ts
```diff
@@ -130,6 +128,18 @@ export const convertDecryptedSeedBufferToString = (decryptedSeed: Buffer): strin
   return `${seedHex}++jolo`
 }
 
+export const processAnchoredElemDidSeed = (parsedSeed: ParseDecryptedSeedResult, did: string) => {
```
inga-affinidi, line 131: The name of this method is misleading, something like `extendSeedWithDid` would be more appropriate.

common-libs/common/src/shared/seedTools.ts
```diff
@@ -130,6 +128,18 @@ export const convertDecryptedSeedBufferToString = (decryptedSeed: Buffer): strin
...
+  const { seed, externalKeys } = parsedSeed
+  const seedWithMethod = `${seed.toString('hex')}++${ELEM_ANCHORED_DID_METHOD}`
+  const additionalData = {
+    ...(externalKeys && { [EXTERNAL_KEYS_KEY]: externalKeys }),
+    [METADATA_KEY]: { anchoredDid: did },
```
inga-affinidi, line 136: Please do not throw the existing metadata and other fields away:
```
const { seed, additionalData: originalAdditionalData } = parsedSeed
...
const additionalData = {
  ...originalAdditionalData,
  [METADATA_KEY]: {
    ...originalAdditionalData?.[METADATA_KEY],
    anchoredDid: did,
  },
}
```

Right now we don't have any other fields in additional data, and metadata does not have any other keys, but this might change in the future, and this method should add anchored did to the other data, not replace all the other data with anchored did and external keys.

sdk/core/src/services/registeringHandler.ts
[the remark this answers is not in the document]
inga-affinidi: OK, I'll do that.

sdk/core/src/services/registeringHandler.ts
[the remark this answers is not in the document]
inga-affinidi: So `acceptDid` method of registry service accepts one `did` in request body and returns entirely different `did` in response body?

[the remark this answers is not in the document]
anton-iskryzhytskyi: yes, and more, then you can resolve whole body of did document by short did

sdk/core/src/services/registeringHandler.ts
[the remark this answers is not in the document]
anton-iskryzhytskyi: From what I know, this part was already implemented on registry side, and it is sidetree responsibility.

sdk/core/test/integration/NetworkMember.test.ts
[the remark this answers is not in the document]
anton-iskryzhytskyi: I met [this issue with cyclic dependency](https://lemmatree.slack.com/archives/C025J7PEG80/p1632406590237700), to omit spending time I think we can add new tests after cyclic dep will be fixed.

sdk/core/src/services/registeringHandler.ts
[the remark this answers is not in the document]
inga-affinidi: I've just carefully looked at the code again, and no, it does not answer even that first part of the question.

`register` does indeed call `anchorDid` with `didDocument`; however, for `elem-anchored`, `anchorDid` does not use `didDocument` in any way (`putDocumentInIpfs` is only called for `jolo`), and only does (after ignoring jolo-specific code and inlining `getAnchoringParams`):

```
export const anchorDid = async (
  api: RegistryApiService,
  encryptedSeed: string,
  password: string,
  didDocument: any,
  anchoredDidElem: boolean = false,
  nonce: number = 0,
): Promise<{ did: string }> => {
  const keysService = new KeysService(encryptedSeed, password)
  const { seed, didMethod } = keysService.decryptSeed()

  const didService = DidDocumentService.createDidDocumentService(keysService)
  const did = didService.getMyDid()

  return { did, didDocumentAddress: '', ethereumPublicKeyHex: '', transactionSignatureJson: '' }

  const response = await api.anchorDid({
    did,
    didDocumentAddress: '',
    ethereumPublicKeyHex: '',
    transactionSignatureJson: '',
    nonce,
    anchoredDidElem,
  })

  return { did: response.body.did }
}
```

So both questions still stand:

1. When, after successful registration, there is an attempt to `buildDidDocument` (e.g. in `Affinity.prototype.signCredential`) for an user with `didMethod: 'elem-anchored'`, `ElemAnchoredDidDocument` attempts to resolve anchored did from seed metadata; did resolver in turn calls `resolveDid` of `RegistryApiService`. How will registry service know what document to return for the anchored did if the original did document seems to be never saved in the first place?
2. The stated objective of FTL-1670 was to move document generation out of SDK to registry service, but SDK still concerns itself with generating documents even for `elem-anchored` (during `register`), is this a problem with this PR or with ticket description?

common-libs/common/src/shared/seedTools.ts
```diff
@@ -128,12 +128,12 @@ export const convertDecryptedSeedBufferToString = (decryptedSeed: Buffer): strin
   return `${seedHex}++jolo`
 }
 
-export const processAnchoredElemDidSeed = (parsedSeed: ParseDecryptedSeedResult, did: string) => {
-  const { seed, externalKeys } = parsedSeed
+export const extendSeedWithDid = (decryptedFullSeed: string, did: string) => {
+  const { seed, metadata, externalKeys } = parseDecryptedSeed(decryptedFullSeed)
```
inga-affinidi, line 132: In theory, `additionalData` can contain other things beyond `metadata` and `externalKeys`, and here you are throwing it away.

(This was already mentioned in the earlier comment)

anton-iskryzhytskyi: I don't like this approach, when it will contain smth we can easily modify this code. My position - it can be as is

anton-iskryzhytskyi: For me it looks like https://stackify.com/premature-optimization-evil/

sdk/core/src/services/registeringHandler.ts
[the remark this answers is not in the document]
anton-iskryzhytskyi: @DenisPopov15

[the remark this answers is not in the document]
anton-iskryzhytskyi: @inga-affinidi https://replika.atlassian.net/wiki/spaces/NETCORE/pages/831947216/State+of+SideTree+and+Affinidi+registry+17.06.2021

common-libs/common/src/shared/seedTools.ts
inga-affinidi: It is not a premature optimization (and premature optimization refers to _performance_ optimization).

You can write "add metadata.anchoredDid field to the existing object, no matter what else it contains".
Or you can write "extract externalKeys and metadata from the existing object, add anchoredDid field to metadata, combine externalKeys and metadata into a new object".

The former is simpler, does not require knowledge of how externalKeys interact with additionalData, and future-proof.
The latter requires this function to know about externalKeys, and can cause subtle bugs in the future when it will erase everything else that `additionalData` can hold by then.

sdk/core/src/services/registeringHandler.ts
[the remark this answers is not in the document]
inga-affinidi: @anton-iskryzhytskyi , that's a long page, and I don't see answers to my questions there.

DenisPopov15 (dismissed):

sdk/core/src/services/registeringHandler.ts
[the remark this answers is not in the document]
inga-affinidi: (done)

common-libs/common/src/shared/seedTools.ts
anton-iskryzhytskyi: I mean, it is just time-wasting as and premature optimization - not more. How smth can be changed to additionalData? Only one way - open this file and add some changes, then add tests for that. 
https://levelup.gitconnected.com/the-problem-you-solve-is-more-important-than-the-code-you-write-d0e5493132c6

sdk/core/src/services/registeringHandler.ts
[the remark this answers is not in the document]
anton-iskryzhytskyi: Including tests errors, I expected that, and again it requires changing other files => make PR bigger => harder to review and finding consensus.

common-libs/common/src/shared/seedTools.ts
inga-affinidi: And the problem this function solving, as is clear from its name, is "add anchored did to metadata". Not "add anchored did to metadata _and_ remove everything that's not external keys or metadata from additional data".

sdk/core/src/services/registeringHandler.ts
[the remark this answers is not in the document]
inga-affinidi: You said that you face the same test errors in your integration tests (which are not yet pushed).
This will be solved once we merge #149 which will happen any minute now.

> again it requires changing other files

Unfortunately git didn't detect that, but the main change was moving almost all of `ElemDidDocument` to `ElemDidDocumentBuilder`. Although now that I think of it, I see a way of fixing it.

> make PR bigger => harder to review

It makes `registeringHandler` much easier to review.

sdk/core/src/services/registeringHandler.ts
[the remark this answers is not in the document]
inga-affinidi: These questions are very important (more important than everything else we've been discussing in this PR) and for me they are blockers.

If I cannot see whether (and how) this solves the original ticket, then some other developers will probably also have the same issue.

If the ticket description is incorrect, it should be updated.

Elsewhere in this PR we were discussing implementation details. Here we are discussing what are we trying to implement.

sdk/core/src/services/registeringHandler.ts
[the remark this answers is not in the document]
inga-affinidi: @anton-iskryzhytskyi , replaced the last commit, now git realizes that `ElemDidDocumentBuilder` is just renamed `ElemDidDocument` and there are less changes.

common-libs/common/src/shared/seedTools.ts
anton-iskryzhytskyi: There are nothing that's not external keys or metadata in additional data

sdk/core/src/services/registeringHandler.ts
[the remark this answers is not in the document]
anton-iskryzhytskyi: @inga-affinidi, with @DenisPopov15 we extend the issue description to make it more clear
So AC of the task now are:
1. Client pass to a registry service map of public keys per operations required by create did:elem operation 
2. new unanchored did returned to a client and get used by SDK as a instance did
3. this did will be anchored in short term according to sidetree protocol

In the code implementation:
1. the map of public keys - did:elem with initial state, it is build on sdk and passed to registry (I'm not sure why this approach was chosen, but it is looks easy for me)
2. anchore-did endpoint of registry returns new did:elem built with sidetree protocol
3. according to sidetree protocol it is stored in mongodb, then by specific interval set of unanchored dids are anchored according to sidetree protocol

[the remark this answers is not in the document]
inga-affinidi: So just to make sure:

* We are building initial document in SDK;
* From this initial document we are building initial did address in SDK; this address will contain all the needed information (including public keys);
* We submit this initial did address to registry service
* Registry service creates a new document from that address
* And new did address from that new document
* Registry service stores that new document for that new address somewhere (mongodb etc)
* Registry service returns that new did address
* We add that new did address to the seed
* And discard the initial document and the initial did address
* New did address can be used later to retrieve stored document from the registry service. (This stored document was generated by registry service and is not related to the one initially generated by SDK).

Correct?

common-libs/common/src/shared/seedTools.ts
inga-affinidi: At this moment, yes. But `extendSeedWithDid` does not have to know about that.

sdk/core/src/services/registeringHandler.ts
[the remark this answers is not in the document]
anton-iskryzhytskyi: yes. Also, mongodb is a temporary holder, as in a short time sidetree implementation will anchor this did in the blockchain (for now it is ropsten testnet). Why we need mongodb then? Due to on mainnet of eth time to add trx to blockchain can take up to several hours.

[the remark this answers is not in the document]
inga-affinidi: OK, I'm also adding this as a comment for entire PR, so that other developers will be able to refer to it in the future.

inga-affinidi:
elem-anchored registration logic in this PR:

* We are building initial document in SDK;
* From this initial document we are building initial did address in SDK; this address will contain all the needed information (including public keys);
* We submit this initial did address to registry service
* Registry service creates a new document from that address
* And new did address from that new document
* Registry service stores that new document for that new address somewhere (mongodb etc)
* Registry service returns that new did address
* We add that new did address to the seed
* And discard the initial document and the initial did address
* New did address can be used later to retrieve stored document from the registry service. (This stored document was generated by registry service and is not related to the one initially generated by SDK).

common-libs/internal-api-clients/package-lock.json
```diff
@@ -10,6 +10,7 @@
       "license": "ISC",
       "dependencies": {
         "@affinidi/affinidi-did-auth-lib": "^1.1.0",
+        "@affinidi/tools-common": "^1.0.0-beta.1",
```
inga-affinidi, line 13: Please do not use `npm install` in individual projects directories; use `npm run bootstrap` in the root monorepo directory instead.

There should be no dependencies on monorepo packages in `package-lock.json` (except for implicit ones brought with `affinidi-did-auth-lib`)

sdk/core/README.md
```diff
@@ -124,11 +124,20 @@ const wallet = AffinidiWallet.openWalletByEncryptedSeed(options, encryptedSeed,
 
 ### Create a new wallet
 
+'elem-anchored' did method returns did:elem, and it is anchored with [sidetree](https://identity.foundation/sidetree/spec/) in [ropsten testnet](https://ropsten.etherscan.io/)  
+note: elem-anchored doesn't support external keys due to Sidetree protocol limitations
```
inga-affinidi, line 128: Do we want to add this check into `ElemAnchoredDidService.buildDidDocumentForRegister`?

inga-affinidi: OK, just noticed this discussion https://lemmatree.slack.com/archives/C025J7PEG80/p1632996281380700?thread_ts=1632996163.380500&cid=C025J7PEG80

inga-affinidi (approved):

sdk/core/README.md
anton-iskryzhytskyi: According to @DenisPopov15, as it is implemented on affinity-registry side I don't see any reasons, and more, if the sidetree standard will change and will allow more space it means we need to keep this moment in mind for this codebase

YuriiSkrypnyk (approved):

PR #152, opened 2021-10-11
[FTL-1999] - Prepare new SDK version to prepare the migration (phase 2).

oleksiipiliugin:
FTL-1999
-  Added `bloom-vault` -> `affinidi-vault` migration related logic.


sdk/core/src/migration/index.ts
```diff
@@ -0,0 +1,162 @@
...
+  }
+
+  /**
+   * Send list of users encrypted VCs stored on `bloom-vault` to the `vault-migration-service` to start migration process to the `affinidi-vault`
+   * @param {Array<vcMigrationList>} vcList
+   */
+  private async migrateCredentials(vcList: VaultCredential[]): Promise<void> {
+    const token = await this.getAuth()
+    const url = `${this.baseUrl}/api/v1/migrateCredentials`
+    return this.api.execute(
```
cosminm-affnd, line 150: We need to handle failures here, let's at least add one or two retries if the call fails.

On the broader scope how will we handle failed chunks assuming all retries have failed? Is it possible to retry the whole migration process at a later date?

sdk/core/src/services/BloomVaultStorageService.ts
```diff
@@ -233,15 +242,32 @@ export default class BloomVaultStorageService {
...
+    try {
+      migration = await this._migrationHelper.getMigrationStatus()
+    } catch (err) {
+      console.log('@@@Vault-migration-service migration status call ends with error: ', err)
+    }
 
-    const credentials = await this._fetchAllDecryptedCredentials(accessToken, storageRegion)
+    if (migration.status === 'done') {
+      return []
+    } else {
```
cosminm-affnd, line 254: minor: no need for an else statement here it makes the code harder to read. See https://medium.com/@matryer/line-of-sight-in-code-186dd7cdea88 for the reasoning.

cosminm-affnd (dismissed):

sdk/core/src/services/BloomVaultStorageService.ts
oleksiipiliugin: ok, will do. thanks

sdk/core/src/services/BloomVaultStorageService.ts
```diff
@@ -233,15 +242,32 @@ export default class BloomVaultStorageService {
...
+      const accessToken = await this._authorizeVcVault(storageRegion)
 
-    if (!types) {
-      return credentials
-    }
+      const credentials = await this._fetchAllDecryptedCredentials(accessToken, storageRegion)
```
YuriiSkrypnyk, line 257: probably we may need extend `_fetchAllDecryptedCredentials` to return `id` from bloom vault along with credential for us to avoid putting same credentials into queue in case of retry

sdk/core/src/migration/index.ts
YuriiSkrypnyk: @cosminm-affnd it will retry each time user reading VCs if migration not done. So in worst case when some VCs didn't get to the queue we will be sending all bloom VCs to queue again. [Here](https://replika.atlassian.net/browse/FTL-2109) is described migration status schema and with `CredentialsInQueue` we will avoid duplications in queue with our approach.

sdk/core/src/migration/index.ts
```diff
@@ -0,0 +1,162 @@
...
+    this.didAuthService = new DidAuthService(didAuthAdapter, this.apiKey, this.baseUrl)
+    this.didAuthAdapter = didAuthAdapter
+    this.keysService = keysService
+    this.platformCryptographyTools = platformCryptographyTools
+    this.apiKey = apiKey
+    this.api = new ApiService(this.baseUrl, {
+      Accept: 'application/json',
+      'Api-Key': apiKey,
+      'Content-Type': 'application/json',
+    })
```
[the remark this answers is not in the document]
oleksiipiliugin: fixed

sdk/core/src/services/BloomVaultStorageService.ts
oleksiipiliugin: fixed

sdk/core/package-lock.json
```diff
@@ -10,8 +10,13 @@
       "license": "ISC",
       "dependencies": {
         "@affinidi/affinity-metrics-lib": "^0.0.25",
+        "@affinidi/common": "^2.0.0-beta.9",
```
inga-affinidi, line 13: Please never use `npm install` inside project directories; only use `npm run bootstrap` from the root monorepo directory.

sdk/core/src/migration/credentials/index.ts
```diff
@@ -0,0 +1,163 @@
...
+    }
+  }
+
+  /**
+   * Gets migration status for user with given token
+   */
+  async getMigrationStatus(): Promise<{ status: string }> {
+    const token = await this.getAuth()
+    const url = 'api/v1/migrationStatus'
+    return this.api.execute(
```
inga-affinidi, line 134: Please just add a new service for that into `internal-api-clients` instead of reimplementing all the base functionality such as manual `fetch`ing and did-auth.

It looks like most of the code changes in this PR is just a reimplementation of generic functionality that is already implemented in `internal-api-clients`.

inga-affinidi (dismissed):

sdk/core/.env.example
```diff
@@ -1,3 +1,4 @@
 TEST_SECRETS={}
 TESTMAIL_API_KEY=
 TESTMAIL_NS=
+DOES_MIGRATION_STARTED=false
```
YuriiSkrypnyk, line 4: we don't need this in env

sdk/core/src/migration/credentials/index.ts
```diff
@@ -0,0 +1,174 @@
...
+    this.apiKey = apiKey
+    this.didAuthService = new DidAuthService(didAuthAdapter, this.apiKey, this.baseUrl)
+    this.keysService = keysService
+    this.platformCryptographyTools = platformCryptographyTools
+    this.api = new ApiService(this.baseUrl, {
+      'Api-Key': apiKey,
+      'X-SDK-Version': version,
+    })
+    this.bloomDid = bloomDid
+    this.tokenRequestTime = Date.now()
```
YuriiSkrypnyk, line 53: I believe token request time should not be defined at constructor

sdk/core/src/services/BloomVaultStorageService.ts
```diff
@@ -61,6 +66,13 @@ export default class BloomVaultStorageService {
       accessApiKey: options.accessApiKey,
       sdkVersion: extractSDKVersion(),
     })
+    this._migrationHelper = new MigrationHelper(
+      options.didAuthAdapter,
+      options.accessApiKey,
+      this._keysService,
+      this._platformCryptographyTools,
+      this.didEth,
```
YuriiSkrypnyk, line 74: you don't have didEth at this moment

sdk/core/src/services/BloomVaultStorageService.ts
```diff
@@ -86,13 +98,15 @@ export default class BloomVaultStorageService {
   /* istanbul ignore next: private function */
   private async _authorizeVcVault(storageRegion: string) {
     const { addressHex, privateKeyHex } = this._getVaultKeys()
-    const didEth = `did:ethr:0x${addressHex}`
+    if (!this.didEth) {
+      this.didEth = `did:ethr:0x${addressHex}`
+    }
```
YuriiSkrypnyk, line 103: we can move this logic to getter

sdk/core/src/services/BloomVaultStorageService.ts
```diff
@@ -233,10 +247,35 @@ export default class BloomVaultStorageService {
   }
 
   public async searchCredentials(storageRegion: string, types?: string[][]): Promise<any[]> {
+    const doesMigrationStarted = await this._migrationHelper.doesMigrationStarted()
+    process.env.DOES_MIGRATION_STARTED = String(doesMigrationStarted)
```
YuriiSkrypnyk, line 251: don't need manipulate with env variable, just operate with `doesMigrationStarted`

sdk/core/src/services/BloomVaultStorageService.ts
```diff
@@ -233,10 +247,35 @@ export default class BloomVaultStorageService {
   }
 
   public async searchCredentials(storageRegion: string, types?: string[][]): Promise<any[]> {
+    const doesMigrationStarted = await this._migrationHelper.doesMigrationStarted()
```
YuriiSkrypnyk, line 250: try-catch here to avoid error and continue processing get VCs. `doesMigrationStarted` should be false by default

sdk/core/src/migration/credentials/index.ts
```diff
@@ -0,0 +1,174 @@
...
+    for (let i = 0, L = encryptedVCs.length; i < L; i += chunk) {
+      const chunkedVCList = encryptedVCs.slice(i, i + chunk)
+      await this.migrateCredentials(chunkedVCList)
+    }
+  }
+
+  /**
+   * Gets migration status for user with given token
+   */
+  async getMigrationStatus(): Promise<{ status: string }> {
```
YuriiSkrypnyk, line 134: returns boolean

sdk/core/src/services/BloomVaultStorageService.ts
```diff
@@ -233,10 +247,35 @@ export default class BloomVaultStorageService {
...
+    process.env.DOES_MIGRATION_STARTED = String(doesMigrationStarted)
+    if (process.env.DOES_MIGRATION_STARTED === 'true') {
+      let migration: any = { status: undefined }
+      try {
+        migration = await this._migrationHelper.getMigrationStatus()
+      } catch (err) {
+        console.log('@@@Vault-migration-service migration status call ends with error: ', err)
```
YuriiSkrypnyk, line 257: make logs readable without `@@@`

sdk/core/src/migration/credentials/index.ts
```diff
@@ -0,0 +1,174 @@
...
+  }
+
+  /**
+   * Send list of users encrypted VCs stored on `bloom-vault` to the `vault-migration-service` to start migration process to the `affinidi-vault`
+   * @param {Array<vcMigrationList>} vcList
+   */
+  async migrateCredentials(vcList: vcMigrationList[]): Promise<void> {
+    const token = await this.getAuth()
+    const url = 'api/v1/migrate/credentials'
+    return this.api.execute(
```
YuriiSkrypnyk, line 162: lets move try-catch to migration helper code

anton-iskryzhytskyi (commented):
@oleksiipiliugin overall looks good except small things
Only one question about the fetch

sdk/core/src/migration/credentials/index.ts
```diff
@@ -0,0 +1,206 @@
...
+        url,
+        {},
+        {
+          Authorization: `Bearer ${token}`,
+        },
+      )
+      if (response) migrationDone = 'yes'
+      else migrationDone = 'no'
+    } catch (err) {
+      console.log('Vault-migration-service migration status check call ends with error: ', err)
```
DenisPopov15, line 156: @oleksiipiliugin Lets do console.error ?

DenisPopov15 (approved):

sdk/core/src/migration/credentials/index.ts
```diff
@@ -0,0 +1,208 @@
...
+      const response = await this.api.execute(
+        'GET',
+        url,
+        {},
+        {
+          Authorization: `Bearer ${token}`,
+        },
+      )
+      if (response) migrationDone = 'yes'
+      else migrationDone = 'no'
```
YuriiSkrypnyk, line 154: this endpoint returns boolean in response in body, so just return result and throw an error in case of error

sdk/core/test/unit/services/BloomVaultStorageService.test.ts
```diff
@@ -56,16 +61,77 @@ describe('BloomVaultStorageService', () => {
...
+      .get('/data/0/99')
+      .reply(200, [
+        { id: 0, cyphertext: JSON.stringify(signedCredential) },
+        { id: 1, cyphertext: JSON.stringify({ ...signedCredential, type: ['type1'] }) },
+      ])
+    nock(bloomVaultUrl, { reqheaders }).get('/data/100/199').reply(200, [])
+    const stub = sinon.stub(MigrationHelper.prototype, 'runMigration').resolves()
+    nock(VAULT_MIGRATION_SERVICE_URL, { reqheaders }).get('/migration/started').reply(200, 'true')
+    const service = createBloomStorageService()
+    nock(VAULT_MIGRATION_SERVICE_URL, { reqheaders }).get(`/migration/done/${service.didEthr}`).reply(200, 'false')
```
YuriiSkrypnyk, line 107: probably here will be good to mock http call to POST `/migration/credentials` instead of stubbing it

YuriiSkrypnyk (commented):
Overall looks good, 2 minor comments only

anton-iskryzhytskyi (approved):

YuriiSkrypnyk (approved):

PR #155, opened 2021-10-13
[FTL-1954] Reorganize packages

inga-affinidi:
did-auth-lib is now a part of monorepo; fixed cyclical dependency `common` -> `internal-api-clients` -> `did-auth-lib` -> `common`


common-libs/tools-common/src/services/JwtService.ts
```diff
@@ -1,32 +1,35 @@
...
 import encode from 'base64url'
 import { randomBytes } from '../shared/randomBytes'
 
-import { DEFAULT_JWT_EXPIRY_MS } from '../_defaultConfig'
+const JwtService = {
+  DEFAULT_JWT_EXPIRY_MS: 10 * 60 * 1000,
 
-export default class JwtService {
-  static getDidFromToken(jwt: string) {
+  keyIdToDid: (keyId: string) => {
```
DenisPopov15, line 8: @inga-affinidi why this logic is moved there? This related logic how to get did from keyId its about didDocument and not JWT really

inga-affinidi: Maybe I misunderstood things.
`JwtService` uses it (e.g. in `buildJWTInteractionToken`), and it is totally independent from the rest of affinidi code, so I thought it should belong here.

(Originally I tried to break `did-auth-lib -> common` dependency, replacing it with `did-auth-lib -> tools-common`, before I realized I should break `internal-api-clients -> did-auth-lib` instead)

common-libs/internal-api-clients/src/helpers/didAuthClientWrapper.ts
```diff
@@ -1,10 +1,15 @@
 import { ThisData as BasicThisData, ClientOptions, createThisData } from './client'
-import { DidAuthAdapter } from './DidAuthAdapter'
 import { DidAuthSession } from './DidAuthManager'
 import { mapFunctions } from './mapFunctions'
 
+export type DidAuthAdapterType = {
+  readonly did: string
+  createDidAuthResponseToken(didAuthRequestToken: string): Promise<string>
```
Yengas, line 7: if we change this method to return `Promise<T>` where `T` is an interface that has `isExpiredAt(now: number)` and `toString(): string` methods. It could simplify the code in `common-libs/internal-api-clients/src/helpers/DidAuthManager.ts`

PR #158, opened 2021-10-18
[FTL-2274] Fix keyId handling

inga-affinidi:
Context: https://lemmatree.slack.com/archives/C025J7PEG80/p1634262053329500

### DIFF
--- a/sdk/core/src/services/anchoringHandler.ts
+++ b/sdk/core/src/services/anchoringHandler.ts
@@ -16,7 +16,7 @@ type AnchoringParams = {
   }
 }
 
-const getPreparedJoloParams = async ({ registry, keysService, nonce, additionalJoloParams }: AnchoringParams) => {
+const computePreparedJoloParams = async ({ registry, keysService, nonce, additionalJoloParams }: AnchoringParams) => {
   if (!additionalJoloParams) {
     throw new Error('missing jolo params')
   }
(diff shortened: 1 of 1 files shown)


sdk/core/src/services/anchoringHandler.ts
```diff
@@ -1,58 +1,68 @@
...
+  nonce: number
+  anchoredDidElem: boolean
+  additionalJoloParams?: {
+    didDocument: any
+    seedHex: string
+  }
+}
 
-    return { did, didDocumentAddress: '', ethereumPublicKeyHex: '', transactionSignatureJson: '' }
+const getPreparedJoloParams = async ({ registry, keysService, nonce, additionalJoloParams }: AnchoringParams) => {
```
anton-iskryzhytskyi, line 19: it is not only gets params but also make some processing around them, let's back what we consider in previous PR related to this https://github.com/affinityproject/affinidi-core-sdk/pull/150#discussion_r717397071

sdk/core/src/CommonNetworkMember/BaseNetworkMember.ts
```diff
@@ -221,8 +222,22 @@ export abstract class BaseNetworkMember {
     nonce: number,
     { basicOptions: { registryUrl }, accessApiKey }: ParsedOptions,
   ) {
-    const api = new RegistryApiService({ registryUrl, accessApiKey, sdkVersion: extractSDKVersion() })
-    return anchorDid(api, encryptedSeed, password, didDocument, false, nonce)
+    const registry = new RegistryApiService({ registryUrl, accessApiKey, sdkVersion: extractSDKVersion() })
+    const keysService = new KeysService(encryptedSeed, password)
+    const { seed, didMethod } = keysService.decryptSeed()
+    const didService = DidDocumentService.createDidDocumentService(keysService)
```
anton-iskryzhytskyi, line 228: the previous variant was shorter, passing did is required now?

sdk/core/src/services/anchoringHandler.ts
inga-affinidi: Ultimately, it returns params used by `anchorDid` method of registry service.

I agree, `get` is not the best verb here.

PR #159, opened 2021-10-18
fix: FTL-2256 fab uses register method returns wallet with resolvable did

anton-iskryzhytskyi:
All old implementation that create common network member returns it with resolvable did (e.g. for old elem with initial state), 
 but new implementation of factory method "createWallet" used method registry that could return not resolvable did (for old elem method - short version without initial state, which obviously can't be resolved)
This PR provides a fix and bunch of int tests

### DIFF
--- a/sdk/core/test/integration/NetworkMember.test.ts
+++ b/sdk/core/test/integration/NetworkMember.test.ts
@@ -20,6 +20,7 @@ import { generateCredentials } from '../helpers/generateCredentials'
 import { SignedCredential } from '../../src/dto'
 import { testPlatformTools } from '../helpers/testPlatformTools'
 import { RegistryApiService } from '@affinidi/internal-api-clients'
+import { DidMethod, SdkOptions } from '../../dist'
 
 const {
   PASSWORD,
@@ -1106,7 +1107,7 @@ describe('CommonNetworkMember', () => {
 
     const resolvedDidDocument = await commonNetworkMember.resolveDid(commonNetworkMember.did)
 
-    expect(resolvedDidDocument.id).equal(commonNetworkMember.didDocumentKeyId.split('#')[0])
+    expect(commonNetworkMember.did.includes(resolvedDidDocument.id)).to.be.eql(true)
     expect(resolvedDidDocument.publicKey.length).to.be.gte(1)
   })
 
@@ -1121,7 +1122,7 @@ describe('CommonNetworkMember', () => {
 
     const resolvedDidDocument = await commonNetworkMember.resolveDid(commonNetworkMember.did)
 
-    expect(resolvedDidDocument.id).equal(commonNetworkMember.didDocumentKeyId.split('#')[0])
+    expect(commonNetworkMember.did.includes(resolvedDidDocument.id)).to.be.eql(true)
     expect(resolvedDidDocument.publicKey.length).to.be.gte(1)
   })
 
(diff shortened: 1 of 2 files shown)


sdk/core/test/integration/NetworkMember.test.ts
```diff
@@ -1094,4 +1094,49 @@ describe('CommonNetworkMember', () => {
...
+      {
+        ...options,
+        didMethod: 'elem',
+      },
+      password,
+    )
+
+    const resolvedDidDocument = await commonNetworkMember.resolveDid(commonNetworkMember.did)
+
+    expect(resolvedDidDocument.id).equal(commonNetworkMember.didDocumentKeyId.split('#')[0])
```
inga-affinidi, line 1109: Can't we just compare it to `commonNetworkMember.did`?

Otherwise these tests will rely on implementation detail that `didDocumentKeyId` is combined of the did you're looking for, plus `#`, plus key name. And this detail changed just recently.

sdk/core/test/integration/NetworkMember.test.ts
```diff
@@ -1094,4 +1094,49 @@ describe('CommonNetworkMember', () => {
     const commonNetworkMember = new AffinidiWallet(password, encryptedSeedElem, options)
     await commonNetworkMember.deleteAllCredentials()
   })
+
+  it('#createWallet should return wallet with resolvable did (elem)', async () => {
```
inga-affinidi, line 1098: I'd also suggest having yet another test: `createWallet`, then `openWalletByEncryptedSeed`, and make sure the two return the same did.

sdk/core/test/integration/NetworkMember.test.ts
```diff
@@ -1094,4 +1094,49 @@ describe('CommonNetworkMember', () => {
...
+    const commonNetworkMember = await AffinidiWallet.createWallet(
+      {
+        ...options,
+        didMethod: 'elem',
+      },
+      password,
+    )
+
+    const resolvedDidDocument = await commonNetworkMember.resolveDid(commonNetworkMember.did)
```
inga-affinidi, line 1107: Do I understand it right that this call would have failed in the current main branch?

inga-affinidi (dismissed):

sdk/core/test/integration/NetworkMember.test.ts
inga-affinidi: > as for old elem with initial state the value of commonNetworkMember.did is not the same as resolvedDidDocumentkeyId

So `resolveDid` returns document with a id different from the did requested?

anton-iskryzhytskyi: it depends on what do you mean when saying "different"

they are different as strings  for old elem
id: _did:elem:EiCLgnTdCKKOoT_nxV2gDbfA0aw7AX4f5vGkZFsOuBX9DQ_
did: _did:elem:EiCLgnTdCKKOoT_nxV2gDbfA0aw7AX4f5vGkZFsOuBX9DQ_;elem:initial-state=eyJwcm90ZWN0ZWQiOiJleUp2Y0dWeVlYUnBiMjRpT2lKamNtVmhkR1VpTENKcmFXUWlPaUlqY0hKcGJXRnllU0lzSW1Gc1p5STZJa1ZUTWpVMlN5SjkiLCJwYXlsb2FkIjoiZXlKQVkyOXVkR1Y0ZENJNkltaDBkSEJ6T2k4dmR6TnBaQzV2Y21jdmMyVmpkWEpwZEhrdmRqSWlMQ0p3ZFdKc2FXTkxaWGtpT2x0N0ltbGtJam9pSTNCeWFXMWhjbmtpTENKMWMyRm5aU0k2SW5OcFoyNXBibWNpTENKMGVYQmxJam9pVTJWamNESTFObXN4Vm1WeWFXWnBZMkYwYVc5dVMyVjVNakF4T0NJc0luQjFZbXhwWTB0bGVVaGxlQ0k2SWpBek5EYzNOV1prTlROa01qUXpPR016WWpWaU9UYzBZamRqTkRjME9UWTROV0UyTkRrMk1UVTRZV1ExWVdVMk56Qm1aakZqWXpnNFpHRXdNVFUyTkdVeVpTSjlMSHNpYVdRaU9pSWpjbVZqYjNabGNua2lMQ0oxYzJGblpTSTZJbkpsWTI5MlpYSjVJaXdpZEhsd1pTSTZJbE5sWTNBeU5UWnJNVlpsY21sbWFXTmhkR2x2Ymt0bGVUSXdNVGdpTENKd2RXSnNhV05MWlhsSVpYZ2lPaUl3TWpZek5UVmlPV1l4TVdFM1lXUmlaR05qWlRSaE5EQXhZalUzTlRSbFptVmpNREkzTnpkbU5XSXdNalV6TUdFeE9ETXlNakUxWldNM01qRmhZelpoTkRBaWZWMHNJbUYxZEdobGJuUnBZMkYwYVc5dUlqcGJJaU53Y21sdFlYSjVJbDBzSW1GemMyVnlkR2x2YmsxbGRHaHZaQ0k2V3lJamNISnBiV0Z5ZVNKZGZRIiwic2lnbmF0dXJlIjoia1NKbGE5elNtYlNsUDg1WDU5bHRWaG5IUzFTVm90c0plR0tST0lTM2NHWVUzaHlzZC01ZGhFYVJidUNaN05TZ1ZaeEJFLWNFTjVDb2U5LXhSa0JUYmcifQ

PR #160, opened 2021-10-19
[FTL-2007] Refactor did-auth-lib; accept arbitrary registry endpoint URL

inga-affinidi:


common-libs/did-auth-lib/src/shared/builder.ts
```diff
@@ -0,0 +1,15 @@
+import { JwtService } from '@affinidi/tools-common'
+import { parse } from 'did-resolver'
+
+export const buildResponseJwtObject = async (didAuthRequestToken: string) => {
+  const didAuthRequestTokenDecoded = JwtService.fromJWT(didAuthRequestToken)
+  const jwtType = 'DidAuthResponse'
+  const NOW = Date.now()
```
vitalyrudenko-affinidi, line 7: Why screaming? :smile:

common-libs/did-auth-lib/src/shared/Signer.ts
```diff
@@ -0,0 +1,42 @@
+import { KeysService, KeyVault } from '@affinidi/common'
+import base64url from 'base64url'
+
+type SignerOptions = {
+  did: string
+  keyId: string
+  keyVault: KeyVault
+}
+
+export default class Signer {
```
vitalyrudenko-affinidi, line 10: This sounds very abstract, though it's a bit tied to the "JWT"? Perhaps we can call it "JwtSigner"?

common-libs/did-auth-lib/src/DidAuthService/DidAuthService.ts
```diff
@@ -94,59 +81,33 @@ export default class AffinidiDidAuthService {
...
-    jwtObject.payload.createdAt = NOW
+  private createSigner() {
+    const signerOptions = {
+      did: this._did,
+      keyId: this._keyId,
+      keyVault: this._keyVault,
+    }
 
-    await this.fillSignature(jwtObject)
+    return new Signer(signerOptions)
```
vitalyrudenko-affinidi, line 95: Why not just:
```
return new Signer({
  did: this._did,
  ...
})
```

common-libs/did-auth-lib/src/shared/builder.ts
inga-affinidi: It is not a new code, just moved: https://github.com/affinityproject/affinidi-core-sdk/blob/a6de74683e066e238828cc679bb87c9c2beef631/common-libs/did-auth-lib/src/DidAuthService/DidAuthService.ts#L103

common-libs/did-auth-lib/src/DidAuthService/DidAuthClientService.ts
```diff
@@ -0,0 +1,34 @@
...
+
+    // the token expiration date here is checked with the client clock
+    // this means if client clock diverges from the server clock;
+    //   - if the client clock is ahead of the server; client may allow tokens that are valid for more than 'maxTokenValidityPeriod'
+    //   - if the client clock is behind the server; client may not sign tokens that are valid for less then 'maxTokenValidityPeriod'
+    if (!didAuthRequestToken.exp || didAuthRequestToken.exp > Date.now() + maxTokenValidityPeriod) {
+      throw new Error(
+        `request token can not be valid more than max token validity period of ${maxTokenValidityPeriod}ms`,
+      )
+    }
```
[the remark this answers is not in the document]
inga-affinidi: This is not a new code, just moved: https://github.com/affinityproject/affinidi-core-sdk/blob/a6de74683e066e238828cc679bb87c9c2beef631/common-libs/did-auth-lib/src/DidAuthService/DidAuthService.ts#L122-L126

And it is beyond the scope of this PR to address _every_ code quality issue in the original code.

inga-affinidi:
@anton-iskryzhytskyi 
> One important note, I didn't find a way to pass custom registry url here

Because now you just pass an entire `Affinidi` object to `DidAuthServerService` constructor.
(In the original implementation, registry url was only used to create an instance of `Affinidi`; and in most cases, we already have an `Affinidi` instance when we use did-auth-lib, so it would not make sense to try and create another one, hoping that the methods we're calling will not depend on other options)

PR #172, opened 2021-10-28
[FTL-2043] Add npm lib to store actual urls of affinidi services for all ends

vladmitchenko:
- Added functionality to resolve url by service name and env
- Added actual services
- Added specific urls for affinidi-schema-manager
- Added unit and integration tests


cosminm-affnd (dismissed):

cosminm-affnd (dismissed):

vladmitchenko:
> > Ideally we make our PRs as atomic as possible otherwise we drag too many changes into one batch and make it hard to review. Assuming we've done our due diligence on the API of the library it's unlikely breaking changes will be needed.
> 
> I disagree. We have this monorepo precisely so that we would be able to check how its parts are working together prior to publishing new versions of npm packages separately.
> 
> Otherwise we could just have ten different repositories, one per each package, and not use lerna, hoping that every change we make to the basic project wil actually be the onen needed by its dependent projects.

Added usage of this lib in SDK as separate package
https://github.com/affinityproject/affinidi-core-sdk/pull/176

oleksiipiliugin (dismissed):

anton-iskryzhytskyi (approved):
LGTM

cosminm-affnd (dismissed):

common-libs/url-resolver/src/index.ts
```diff
@@ -0,0 +1,19 @@
+import { defaultTemplate, predefinedTemplates } from './templates'
+import { predefinedUrls } from './urls'
+import { Service } from './services'
+
+function resolveUrl(service: Service, env: string, userTemplate?: string): string {
```
inga-affinidi, line 5: `env` should be of type `'dev' | 'staging' | 'prod'` (or you can make it an enum if you want to)

oleksiipiliugin (approved):

cosminm-affnd (approved):

PR #175, opened 2021-11-01
[FTL-1589] Remove tsdx from core SDK

stephanie-affinidi:


.npmrc
```diff
@@ -0,0 +1 @@
+legacy-peer-deps=true
```
inga-affinidi, line 1: Why do we need this on monorepo level?
(Also why do we need this at all?)

stephanie-affinidi: NPM v7 changed the way peer dependencies work, and there seem to be some incompatibilities with some of the libraries we use with that. Since nodejs LTS comes with NPM v6 (unless you upgrade it manually, which is supported), this issue hasn't been seen before by those using homebrew, but it will rear it's head as soon as nodejs is upgraded.

I'm happy to remove this and let this be addressed in a later PR though.

inga-affinidi: @stephanie-affinidi we are using NPM v7 everywhere.
I still don't understand why do we need this on monorepo level (i.e. on the projects you didn't touch as well)? What problems will we have if we remove this file? Will these problems be solved if we put it into directories of individual projects you've touched in this PR?

stephanie-affinidi: If I check out a clean copy of `origin/master`, and then run `npm install` and then `npm run bootstrap` and then `npm  run lint`, then I get this error:
```
ESLint couldn't find the plugin "@typescript-eslint/eslint-plugin".

(The package "@typescript-eslint/eslint-plugin" was not found when loaded as a Node module from the directory "/Users/stephaniewilde-hobbs/affinidi/affinidi-core-sdk/master/common-libs/did-auth-lib".)

It's likely that the plugin isn't installed correctly. Try reinstalling by running the following:

    npm install @typescript-eslint/eslint-plugin@latest --save-dev

The plugin "@typescript-eslint/eslint-plugin" was referenced from the config file in ".eslintrc.js » @affinidi/eslint-config".
```

I thought that I made this error go away by adding this `.npmrc`, but it doesn't seem to be working any more... I'm very confused at this point cause linting doesn't work even on a clean copy of the codebase...

PR #177, opened 2021-11-03
[FTL-2291] Add and apply usage @affinidi/url-resolver to did-auth-lib and common

vladmitchenko:


common-libs/did-auth-lib/src/DidAuthService/DidAuthService.ts
```diff
@@ -118,15 +119,15 @@ export default class AffinidiDidAuthService {
   ): Promise<string> {
     const cloudWallet = new CloudWalletApiService({
       accessApiKey: apiKey,
-      cloudWalletUrl: `https://cloud-wallet-api.${environment}.affinity-project.org/api/v1/utilities/sign-jwt`,
+      cloudWalletUrl: resolveUrl(Service.CLOUD_WALLET_API, <Env>environment) + '/api/v1/utilities/sign-jwt',
```
inga-affinidi, line 122: It would be better if instead `createDidAuthResponseTokenThroughCloudWallet` just accepted `environment: Env`.

common-libs/did-auth-lib/src/DidAuthService/DidAuthService.ts
```diff
@@ -118,15 +119,15 @@ export default class AffinidiDidAuthService {
...
     })
     const cloudService = new DidAuthCloudService(cloudWallet, cloudWalletAccessToken)
     return cloudService.createDidAuthResponseToken(didAuthRequestToken)
   }
 
   async verifyDidAuthResponseToken(didAuthResponseTokenStr: string, options: VerifierOptions): Promise<boolean> {
     const affinidiOptions = {
-      registryUrl: `https://affinity-registry.${options.environment}.affinity-project.org`,
+      registryUrl: resolveUrl(Service.REGISTRY, <Env>options.environment),
```
inga-affinidi, line 130: Same as above, type of `VerifierOptions.environment` should be `Env` and not `string`.

YuriiSkrypnyk (approved):

anton-iskryzhytskyi (approved):

PR #182, opened 2021-11-12
feat: add undici as http client

vladmitchenko:


anton-iskryzhytskyi (dismissed):
LGTM

cosminm-affnd (approved):

anton-iskryzhytskyi (dismissed):

inga-affinidi:
What is the purpose of adding undici support? What is the associated jira ticket?

vladmitchenko:
> What is the purpose of adding undici support? What is the associated jira ticket?

[FTL-2557](https://lemmatree.atlassian.net/browse/FTL-2557)

anton-iskryzhytskyi (dismissed):

oleksiipiliugin (approved):

common-libs/internal-api-clients/CHANGELOG.md
```diff
@@ -1,3 +1,5 @@
+# release 1.0.0-beta.15 (2021-11-12)
+  * Added `undici` as http client for nodejs version >= 16
```
inga-affinidi, line 2: The reason for adding support for new http client is unclear.
It increases complexity of our code, especially since it is not supported on all platforms.
It provides some performance benefits over axios but http requests performance, but it is not clear that http requests overhead is a bottleneck for SDK performance.
I do not see any clear benefits on the provided benchmarks.
And undici's own benchmarks claim to reduce the overhead of a single HTTP request from 0.3ms to to 0.1ms. Which makes it seem like this 3x performance optimization only covers fractions of a percent of actual SDK latency.
Typical SDK usage scenarios take several seconds to complete, and they might have, _very generously_, maybe 100 requests. So it seems that we can expect undici to reduce the total overhead from 30ms to 10ms _at most_ (which is less than a percent of the total scenario runtime, again, on a very generous estimate), and that's only on a single platform.

inga-affinidi (dismissed):

common-libs/internal-api-clients/CHANGELOG.md
anton-iskryzhytskyi: The problem is only for cloud-wallet, one of the OKR for decentralization team is support more than 10kk requests per day, but once cloud-wallet receive load more then, let's say 40 rps for 2 instances the time for http requests with node-fetch increasing exponentially. https://app.k6.io/runs/1162700

oleksiipiliugin (approved):

common-libs/internal-api-clients/CHANGELOG.md
anton-iskryzhytskyi: Also, according to our tests listed in the ticket the difference becoming big after huge load [FTL-2557](https://lemmatree.atlassian.net/browse/FTL-2557)

anton-iskryzhytskyi (approved):

common-libs/internal-api-clients/CHANGELOG.md
vladmitchenko: Also undici http client is optional and can be enabled with env variables

We really need it, cuz it can help increase cloud-wallet-api performance, and we going to test it on staging, if it won't show good results, we can cancel this change

inga-affinidi: @anton-iskryzhytskyi , this link does not work for me for some reason. Could you please tell me more about the results?

Do I understand it right that:
1. We have an actual scenario where cloud-wallet-api performance is inadequate; and
2. We are absolutely certain it is caused by node-fetch performance issues and not any other bottlenecks; and
3. We are certain that migrating to undici performance of SDK at least in some scenarios will improve significantly?

anton-iskryzhytskyi: 1. yes https://app.k6.io/runs/public/c7a34cf0e98f43dbb4c190deb03514f4 dev before all improvements

PR #183, opened 2021-11-12
Disable normalisation for different type of cognito pool

dmfilipenko-affinidi:
It's a part of effort to make some Cognito pool reusable


sdk/core/src/services/UserManagementService.ts
```diff
@@ -92,9 +92,13 @@ export default class UserManagementService {
     }
   }
 
-  async signUpWithUsernameAndConfirm(username: string, inputPassword: string) {
+  async signUpWithUsernameAndConfirm(
+    username: string,
+    inputPassword: string,
+    shouldDisableNameNormalisation?: boolean,
```
inga-affinidi, line 98: Please make this a constructor option, as we discussed already

sdk/core/src/services/UserManagementService.ts
```diff
@@ -377,9 +395,9 @@ export default class UserManagementService {
     }
   }
 
-  private _buildUserAttributes(login: string) {
+  private _buildUserAttributes(login: string, shouldDisableNameNormalisation: boolean) {
```
inga-affinidi, line 398: but it is an instance method, it does not need this parameter, it can read it from `this`.

sdk/core/src/services/UserManagementService.ts
```diff
@@ -58,6 +59,7 @@ export default class UserManagementService {
   private _cognitoIdentityService
   private _keyStorageApiService
   private _sessionStorageService
+  private _shouldDisableNameNormalisation = false
```
inga-affinidi, line 62: but this default value only applies if you do not assign this field, and we're always assigning it?

sdk/core/src/services/UserManagementService.ts
```diff
@@ -67,6 +69,7 @@ export default class UserManagementService {
     })
     this._cognitoIdentityService = new CognitoIdentityService(options)
     this._sessionStorageService = new SessionStorageService(options.userPoolId)
+    this._shouldDisableNameNormalisation = options.shouldDisableNameNormalisation
```
inga-affinidi, line 72: `?? false`

inga-affinidi (changes requested):

vitalyrudenko-affinidi (approved):

inga-affinidi (approved):

PR #185, opened 2021-11-15
Generate password with special character

dmfilipenko-affinidi:
Some of pools request password with special character

### DIFF
--- a/sdk/core/src/services/UserManagementService.ts
+++ b/sdk/core/src/services/UserManagementService.ts
@@ -26,7 +26,7 @@ import { extractSDKVersion } from '../_helpers'
 const generatePassword = async () => {
   const randomPassword = (await randomBytes(32)).toString('hex')
   // Make first found letter uppercase because hex string doesn't meet password requirements
-  // Also add some special characters to the end of password
+  // Special characters at the end of password to make comply with cognito requirements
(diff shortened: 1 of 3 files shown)


vitalyrudenko-affinidi (approved):

sdk/core/src/services/UserManagementService.ts
```diff
@@ -26,7 +26,8 @@ import { extractSDKVersion } from '../_helpers'
 const generatePassword = async () => {
   const randomPassword = (await randomBytes(32)).toString('hex')
   // Make first found letter uppercase because hex string doesn't meet password requirements
-  return randomPassword.replace(/[a-f]/, 'A')
+  // Special characters at the end of password to make comply with cognito requirements
+  return randomPassword.replace(/[a-f]/, 'A') + '!'
```
inga-affinidi, line 30: We definitely should write some TODO here...
```
// TODO: Actually generate a random password instead of simply replacing first letter of hexadecimal password with 'A' and appending an exclamation mark
```

inga-affinidi (approved):

sdk/core/src/services/UserManagementService.ts
dmfilipenko-affinidi: Good catch. Sure, let's keep it for later

PR #186, opened 2021-11-16
[WIP] FTL-2399 user management lib

dmfilipenko-affinidi:


common-libs/user-management/src/dto.ts
```diff
@@ -0,0 +1,31 @@
+import { IsOptional, IsNumber, IsString } from 'class-validator'
```
inga-affinidi, line 1: But we are not actually doing any validation here? So no need to depend on class-validator.

sdk/core/src/CommonNetworkMember/LegacyNetworkMemberWithFactories.ts
```diff
@@ -6,13 +6,20 @@ import { withDidData } from '../shared/getDidData'
 import { getOptionsFromEnvironment, ParsedOptions } from '../shared/getOptionsFromEnvironment'
 import { ParametersValidator } from '../shared/ParametersValidator'
 import { randomBytes } from '../shared/randomBytes'
-import { validateUsername } from '../shared/validateUsername'
-import UserManagementService from '../services/UserManagementService'
+import { validateUsername, UserManagementService } from '@affinidi/user-management'
 import { StaticDependencies, ConstructorUserData, createKeyManagementService } from './BaseNetworkMember'
 import { LegacyNetworkMember } from './LegacyNetworkMember'
+import { DEFAULT_COGNITO_REGION } from '../_defaultConfig'
+import { KeyStorageApiService } from '@affinidi/internal-api-clients'
```
inga-affinidi, line 13: Please maintain the alphabetic order of dependencies, in this line and in line 9.

(We really should update eslint-config dependency, the newer config should catch things like that)

sdk/core/src/CommonNetworkMember/LegacyNetworkMemberWithFactories.ts
```diff
@@ -6,13 +6,20 @@ import { withDidData } from '../shared/getDidData'
...
+import { extractSDKVersion } from '../_helpers'
 
 const createUserManagementService = ({ basicOptions, accessApiKey }: ParsedOptions) => {
-  return new UserManagementService({ ...basicOptions, accessApiKey })
+  const keyStorageApiService = new KeyStorageApiService({
+    keyStorageUrl: basicOptions.keyStorageUrl,
+    accessApiKey,
+    sdkVersion: extractSDKVersion(),
+  })
+  return new UserManagementService({ ...basicOptions, region: DEFAULT_COGNITO_REGION }, { keyStorageApiService })
```
inga-affinidi, line 22: Looks like it's time to extract `createUserManagementService` into a separate module and avoid all that code duplication.

inga-affinidi (changes requested):
Overall LGTM, besides these minor comments.

common-libs/user-management/src/dto.ts
vitalyrudenko-affinidi: It's copy-pasted from the original code. Should we replace/remove this?

inga-affinidi: We're (slightly) modifying code in this MR anyway, so there is no reason to carry this unneeded dependency into the new package.

common-libs/user-management/package.json
```diff
@@ -0,0 +1,74 @@
+{
+  "name": "@affinidi/user-management",
+  "version": "1.0.0-beta.1",
```
inga-affinidi, line 3: no need in beta suffix imo

inga-affinidi (dismissed):

inga-affinidi (approved):

stephanie-affinidi (approved):

PR #187, opened 2021-11-17
FTL-2364 - test scripts fixed for sdk v6

ezgi-affinidi:


sdk/core/test/helpers/testSecrets.ts
```diff
@@ -24,4 +24,6 @@ export const testSecrets: {
   DEV_API_KEY_HASH: string
   PROD_API_KEY_HASH: string
   STAGING_API_KEY_HASH: string
+  REVOCATION_ENCRYPTED_SEED: string
+  REVOCATION_PASSWORD: string
```
inga-affinidi, line 28: Did we add these to github secrets for CI?
Could it be the reason for tests failing? (Github is glitchy right now and won't let me see why they failed)

ezgi-affinidi: Yes we added, this error in the pipeline didn't appear before with your branch, is it possible to merge v6 first and then fix the test errors? or I can investigate the error.

inga-affinidi: We cannot merge v6 until all tests pass (both manually against prod, and in CI). Or we would have merged it three weeks ago...

inga-affinidi (approved):

vitalyrudenko-affinidi (approved):

PR #193, opened 2021-11-24
feat: ftl-2682 add selective disclosure for nested fields

anton-iskryzhytskyi:


common-libs/common/src/utils/objectUtil.ts
```diff
@@ -0,0 +1,67 @@
...
+  field: string,
+  value: unknown,
+): Record<string, any> => {
+  const keys = Object.keys(target)
+  if (keys.length > 0) {
+    target[field] = value
+    keys.forEach((key) => injectFieldForAllParentRoots(target[key], field, value))
+  }
+
+  return target
```
inga-affinidi, line 50: It could be much easier to understand with lodash:
```
if (!target.keys.length) { // what is the purpose of this condition?
  return {}
}

return {
  mapValues((targetValue) => injectFieldForAllParentRoots(targetValue, field, value))
  [field]: value,
}
```

Bonus point: `target` will no longer be mutated.

Other methods here can also be rewritten in a similar way.

common-libs/common/src/utils/objectUtil.ts
```diff
@@ -0,0 +1,67 @@
...
+  Object.keys(source).forEach((key) => {
+    if (source[key]) {
+      if (!target[key]) Object.assign(target, { [key]: {} })
+      mergeDeep(target[key], source[key])
+    } else {
+      Object.assign(target, { [key]: source[key] })
+    }
+  })
+
+  return mergeDeep(target, ...sources)
```
inga-affinidi, line 66: Looks like recursion is not actually leveraged in this line, and this method can be rewritten as
```
for (const source of sources) {
  // lines 57-64 go here
}

return target
```

And the loop content can be rewritten as
```
const target = {
  ...target,
  mapValues(source, (value, key) => {
    if (!value) {
      return value
    }

    return mergeDeep(target[key] ?? {}, value)
  })
}
```

Or the whole method, without mutating target:
```
return sources.reduce(
  (source, result) => ({
    ...target,
    mapValues(source, (value, key) => {
      if (!value) {
        return value
      }

      return mergeDeep(target[key] ?? {}, value)
    })
  }),
  target
)
```

inga-affinidi: Also, why not just use `merge` method from lodash? (It is a bit different than what is implemented here, but presumably we can change the callers?)

common-libs/common/src/Affinity.ts
```diff
@@ -624,21 +625,21 @@ export class Affinity {
...
-    for (const field of fields) {
-      if (credential.credentialSubject.data[field] === undefined) {
-        throw new Error(`Field "${field}" not a part of credential`)
-      }
-
-      dataFields[field] = {}
+    const validationErrors = validateObjectHasPaths(credential.credentialSubject.data, fields)
+    if (validationErrors) {
+      const firstError = validationErrors[0]
+      throw new Error(`Field "${firstError.field}" not a part of credential`)
```
cosminm-affnd, line 637: Ideally the error would contain all missing fields, not just the first one that's missing. It's not good developer experience if you get errors one by one.

common-libs/common/src/Affinity.ts
```diff
@@ -624,21 +625,21 @@ export class Affinity {
...
     }
 
+    const dataFields = buildObjectSkeletonFromPaths(fields)
```
cosminm-affnd, line 640: What's the advantage of implementing on our own over using an existing library like https://www.npmjs.com/package/json-mask ?

anton-iskryzhytskyi:
This PR is blocked by https://github.com/mattrglobal/node-bbs-signatures/pull/175 , https://github.com/mattrglobal/bbs-signatures/pull/110 and releasing these libs with new versions.

@inga-affinidi @cosminm-affnd tried to omit installing new deps. But I see it is not a common way. So, will follow your recommendations.

common-libs/common/src/Affinity.ts
anton-iskryzhytskyi: I updated syntax to json-mask variant ('/' symbol instead of '.')
I haven't found any useful solution as all of them have specific usage and in case of json-mask it generates specific object, not equal what we need, and it will require **more code** to convert json-mask.compile output with jsonld-signatures-bbs.deriveProof input

anton-iskryzhytskyi:
In this PR I also remove checking of cedentialSubject.data nested fields as according to jsonld context, some fields could be not represented - one VC with jsonld context X can have field 'address'  but for another VC with jsonld context X might not have  field 'address'. In case when for derive proof operation not existing field is revealed, this field in the result segment will have 'null'.

oleksiipiliugin (approved):

YuriiSkrypnyk (approved):

PR #195, opened 2021-11-25
Fix problem of undici with not nodejs platforms (PART 1)

vladmitchenko:


vitalyrudenko-affinidi (approved):

inga-affinidi (changes requested):
This will be fixed as a part of https://lemmatree.atlassian.net/browse/FTL-2702 ; and this PR does not solve a problem with webpack trying to require `undici` anyway.

vitalyrudenko-affinidi:
@inga-affinidi it should. If package is not install, webpack will show a warning - yet will build the project anyway.

anton-iskryzhytskyi:
@vladmitchenko let's close it for now, as we decided to use @inga-affinidi approach

inga-affinidi:
This is solved by #198

PR #197, opened 2021-11-25
changed: bumped version of `@affinidi/wallet-core-sdk`

oleksiipiliugin:
changed: bumped version of `@affinidi/wallet-core-sdk`

### DIFF
--- a/sdk/browser/package-lock.json
+++ b/sdk/browser/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@affinidi/wallet-browser-sdk",
-  "version": "6.0.0",
+  "version": "6.0.1",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "@affinidi/wallet-browser-sdk",
-      "version": "6.0.0",
+      "version": "6.0.1",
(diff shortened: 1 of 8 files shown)


cosminm-affnd (dismissed):

inga-affinidi:
What is the purpose of this PR?

Merging it won't trigger any publishes, because versions of packages themselves are not changed.
Even if they were, clean install of `wallet-browser-sdk` will install the latest version of `wallet-core-sdk` matching `^6.0.0` (and `6.0.2` matches it) anyway.

YuriiSkrypnyk (approved):

YuriiSkrypnyk:
> What is the purpose of this PR?
> 
> Merging it won't trigger any publishes, because versions of packages themselves are not changed. Even if they were, clean install of `wallet-browser-sdk` will install the latest version of `wallet-core-sdk` matching `^6.0.0` (and `6.0.2` matches it) anyway.

Yes, that is correct for clean install, but in case someone already fetched `6.0.0` to be sure that proper version will be installed better pin `6.0.2` for confidence.

cosminm-affnd (approved):

PR #198, opened 2021-11-25
[FTL-2702] Dependency on node-fetch and undici extracted into a separate package

inga-affinidi:
Instead of explicitly using `node-fetch` or `undici` or native `fetch` everywhere, and instead of changing global variables, here we have three packages:
* `platform-fetch`, which does not have any dependencies (not even dev dependencies);
* `platform-fetch-native`, which only has a peer dependency on `platform-fetch`;
* `platform-fetch-node`, which also has that peer dependency, plus dependencies on `node-fetch` and `undici`.

If `fetch` is needed somewhere, it can be used with `import { fetch } from '@affinidi/platform-fetch`

Prior to application startup, it should be configured to use a specific fetch implementation:

```
import { useNativeFetch } from '@affinidi/platform-fetch-native'
useNativeFetch()
```
or
```
import { useNodeFetch } from '@affinidi/platform-fetch-node'
useNodeFetch()
```

Application intended to be run in browsers will never have non-dev dependencies on `platform-fetch-node`, will never require or import `platform-fetch-node`, and therefore there will be no direct or indirect references to `node-fetch` or `undici`, so that webpack will never try to load these.

All our platform-specific SDK packages already invoke `useNodeFetch` or `useNativeFetch` and already have the relevant dependencies, so that client applications can just import `wallet-browser-sdk` as they always did; they won't have `node-fetch` or `undici` in their `node_modules`, and webpack will never try to load these.


sdk/core/src/migration/credentials/ApiService.ts
```diff
@@ -16,6 +16,7 @@ interface FETCH_OPTIONS {
 
 const bodyRequiredMethods = ['POST', 'PATCH', 'PUT']
 
+// TODO: Why do we need this instead of using internal-api-clients?
```
inga-affinidi, line 19: Already asked this question [in original PR](https://github.com/affinityproject/affinidi-core-sdk/pull/152#discussion_r727631511) (where it was marked as "resolved" without any response or changes) and on slack.

This is the only place that uses `node-fetch` directly. This entire class seems to duplicate the functionality of `tools-openapi`, and probably could have been implemented in a similar fashion to everything else in `internal-api-clients`.

So this code probably should be removed, but in order to do that we have to know why it exists in the first place.

platform/fetch-node/package.json
```diff
@@ -0,0 +1,73 @@
...
+  "contributors": [
+    "Roman Brazhnyk <roman@affinity-project.org>"
+  ],
+  "license": "ISC",
+  "bugs": {
+    "url": "https://github.com/affinityproject/affinidi-core-sdk/issues"
+  },
+  "homepage": "https://github.com/affinityproject/affinidi-core-sdk#readme",
+  "dependencies": {
+    "@affinidi/platform-fetch": "^1.0.0",
```
inga-affinidi, line 44: This should be a peer dependency, but lerna does not support peer dependencies.
Hopefully we can fix this once we switch to npm workspaces.

anton-iskryzhytskyi (approved):
Big update, but LGTM
I suggest bumping up versions of sdk/* and common-libs/common|did-auth-lib and then let's merge

stephanie-affinidi:
Why is this PR so big and appears to include so many unrelated changes?

platform/fetch/src/index.ts
```diff
@@ -0,0 +1,20 @@
+import type { RequestInfo, RequestInit, Response } from 'node-fetch'
+import { request as undiciRequest } from 'undici/types/api'
+
+type OptionsType = RequestInit & Parameters<typeof undiciRequest>[1]
+
+type FetchType = (url: RequestInfo, options?: OptionsType) => Promise<Response>
+
+let fetch: FetchType | null = null
+
+export const getFetch = () => {
```
stephanie-affinidi, line 10: Why not wrap fetch like this to keep the fetch API the same:
```js
export function fetch(url: RequestInfo, options?: OptionsType) {
  if (!fetch) {
    throw new Error('fetch is not configured')
  }

  return fetch(url, options)
}
```

inga-affinidi:
@stephanie-affinidi 
> Why is this PR so big and appears to include so many unrelated changes?

I didn't intend for it to become that big, I'll try to split it into two now.

inga-affinidi:
Extracted refactoring of migration service to #199

platform/fetch-native/test/integration/index.test.ts
```diff
@@ -0,0 +1,5 @@
+import 'mocha'
+
+describe.skip('PlatformFetch', () => {
+  // TODO: Write actual tests
```
stephanie-affinidi, line 4: Is this PR still wip? Because it's unlikely this TODO will ever be completed if this is merged as-is. Either this will not ever have tests or it gets merged with them.

platform/fetch-native/src/index.ts
```diff
@@ -0,0 +1,8 @@
+import { configureFetch } from '@affinidi/platform-fetch'
+
+export const useNativeFetch = () => {
+  // TODO: fix the types
```
stephanie-affinidi, line 4: Is this still TODO?

platform/fetch-native/test/unit/index.test.ts
```diff
@@ -0,0 +1,5 @@
+import 'mocha'
+
+describe.skip('PlatformFetch', () => {
+  // TODO: Write actual tests
```
stephanie-affinidi, line 4: ditto

platform/fetch/src/index.ts
```diff
@@ -0,0 +1,20 @@
...
+
+export const fetch: FetchType = (url: RequestInfo, options?: OptionsType) => {
+  if (!fetchImpl) {
+    throw new Error('fetch is not configured')
+  }
+
+  return fetchImpl(url, options)
+}
+
+export const configureFetch = (inputFetch: FetchType) => {
```
stephanie-affinidi, line 18: `setFetchImpl` might be a better name, but I'm happy to approve this with the current name.

platform/fetch/test/integration/index.test.ts
```diff
@@ -0,0 +1,5 @@
+import 'mocha'
+
+describe.skip('PlatformFetch', () => {
+  // TODO: Write actual tests
```
stephanie-affinidi, line 4: ditto

platform/fetch/test/unit/index.test.ts
```diff
@@ -0,0 +1,5 @@
+import 'mocha'
+
+describe.skip('PlatformFetch', () => {
+  // TODO: Write actual tests
```
stephanie-affinidi, line 4: ditto

platform/fetch-native/src/index.ts
inga-affinidi: This was originally intended to be a quick fix for a critical bug, but apparently nobody is interested in it enough, so I'll try to spend some more time on solving all TODOs.

platform/fetch/src/index.ts
inga-affinidi: You're right, it is a better name

platform/fetch-native/test/integration/index.test.ts
inga-affinidi: Tests were not called anyway, removed these files.

platform/fetch/test/integration/index.test.ts
inga-affinidi: removed

platform/fetch/test/unit/index.test.ts
inga-affinidi: removed

platform/fetch-native/test/unit/index.test.ts
inga-affinidi: removed

sdk/core/src/migration/credentials/ApiService.ts
inga-affinidi: Refactored in #199

platform/fetch-node/package.json
```diff
@@ -0,0 +1,58 @@
...
+    "url": "git+https://github.com/affinityproject/affinidi-core-sdk.git",
+    "directory": "sdk/core"
+  },
+  "keywords": [
+    "sdk",
+    "affinity",
+    "network",
+    "did"
+  ],
+  "author": "Denis Popov <denis@slatestudio.com>",
```
dmfilipenko-affinidi, line 29: Can use proper email?

platform/fetch-node/package.json
```diff
@@ -0,0 +1,58 @@
...
+  "contributors": [
+    "Roman Brazhnyk <roman@affinity-project.org>"
```
dmfilipenko-affinidi, line 31: Same here

platform/fetch-node/src/index.ts
```diff
@@ -0,0 +1,20 @@
...
+type FetchType = (url: RequestInfo, init?: RequestInit) => Promise<Response>
+
+export const useNodeFetch = () => {
+  if (process.env.HTTP_CLIENT === 'undici') {
+    setFetchImpl(async (url, options) => {
+      const response = await undiciRequest((url as URL).href ?? url, options)
+
+      return {
+        status: response.statusCode,
+        json: () => response.body.json(),
```
dmfilipenko-affinidi, line 14: Are we only relying on json? 
Also, no `error` method from original response, or it's enough for our needs?

platform/fetch/package.json
```diff
@@ -0,0 +1,56 @@
...
+    "url": "git+https://github.com/affinityproject/affinidi-core-sdk.git",
+    "directory": "sdk/core"
+  },
+  "keywords": [
+    "sdk",
+    "affinity",
+    "network",
+    "did"
+  ],
+  "author": "Denis Popov <denis@slatestudio.com>",
```
dmfilipenko-affinidi, line 29: Lets use correct emails

platform/fetch-node/src/index.ts
stephanie-affinidi: This is the same code that was used for undici before. It's definitely a good idea to improve undici's feature parity with node-fetch, but that appears to be out of scope for this PR.

stephanie-affinidi (approved):

platform/fetch-native/src/index.ts
inga-affinidi: Fixed

platform/fetch-node/src/index.ts
inga-affinidi: There is a full type safety now (no type assertions in fetch implementations). If whatever current implementation has would not be enough for our needs, we would get a compile-time error.

anton-iskryzhytskyi (approved):
LGTM

common-libs/common/src/_baseDocumentLoader/index.ts
```diff
@@ -77,18 +86,7 @@ export const baseDocumentLoader = async (url: string) => {
...
-      res = await fetch(url, {
-        method: 'GET',
-        headers: {
-          Accept: 'application/ld+json, application/json',
-        },
-      })
-    } catch (error) {
-      throw new Error(`URL could not be dereferenced, an error occurred. ${error}`)
-    }
+    const res = await fetchSchema(url)
```
vitalyrudenko-affinidi, line 89: Maybe `fetchDocument`? Or at least `fetchContext`. I don't think "schema" is the correct term here.

common-libs/url-resolver/package.json
```diff
@@ -29,9 +29,9 @@
     "tsc": "tsc -p tsconfig-build.json",
     "lint": "eslint src test",
     "test": "npm run test:unit && npm run test:integration",
-    "test:unit": "NODE_ENV=test mocha -r ts-node/register test/unit/index.test.ts -p ./tsconfig.json --timeout 8000",
-    "test:integration": "NODE_ENV=test mocha -r ts-node/register test/integration/index.test.ts -p ./tsconfig.json --timeout 10000 --exit",
-    "test:coverage": "NODE_ENV=test nyc --reporter=lcov mocha -r ts-node/register test/unit/index.test.ts -p ./tsconfig.json --timeout 8000",
+    "test:unit": "cross-env NODE_ENV=test mocha -r ts-node/register test/unit/index.test.ts -p ./tsconfig.json --timeout 55000",
```
vitalyrudenko-affinidi, line 32: Is 8 seconds is not enough? :D

package.json
```diff
@@ -10,15 +10,15 @@
     "tsc": "lerna run tsc",
     "link": "lerna link",
     "version": "lerna version",
-    "lint": "lerna run --parallel lint",
+    "lint": "lerna run --no-bail --parallel lint",
```
vitalyrudenko-affinidi, line 13: What "--no-bail" is necessary? To show all errors, not just first ones?

sdk/core/package.json
```diff
@@ -63,14 +63,14 @@
...
     "lodash.uniq": "^4.5.0",
-    "node-fetch": "^2.6.0",
     "randombytes": "^2.1.0",
     "secp256k1": "^4.0.2",
     "tslib": "^2.0.0",
     "uuid": "^8.3.1"
   },
   "devDependencies": {
     "@affinidi/eslint-config": "1.0.1",
+    "@affinidi/platform-fetch-node": "^1.0.0",
```
vitalyrudenko-affinidi, line 73: It's dev dependency because it's only used in tests?

vitalyrudenko-affinidi (commented):
Overall looks great!

sdk/core/package.json
anton-iskryzhytskyi: I think yes

platform/fetch/package.json
anton-iskryzhytskyi: Better to create a separate PR to update it everywhere or even change to "author": "Affinidi", (with respect to original authors, but let's highlight we are team not individual devs)

anton-iskryzhytskyi (approved):
LGTM

Let's think about fully replacing authors with "Affinidi" in all packages. [Thread](https://github.com/affinityproject/affinidi-core-sdk/pull/198/files#r763763344)

common-libs/url-resolver/package.json
inga-affinidi: I've just copied commands from other packages in monorepo (my problem with original commands was that they don't run under Windows)

inga-affinidi: It's 55 seconds everywhere in monorepo

package.json
inga-affinidi: Yes, especially considering that tests from the remaining packages will still continue to run, we just won't see their results (and will only be able to detect whether they are still running by checking whether there are some `node` processes)

sdk/core/package.json
inga-affinidi: yes, it is only used in tests (which are run on node)

common-libs/url-resolver/package.json
vitalyrudenko-affinidi: You could've just added "cross-env". IMO timeout for unit tests should stay extremely low.

vitalyrudenko-affinidi (approved):

common-libs/url-resolver/package.json
inga-affinidi: That's another topic.
Easier to have all test commands in all test packages be the same than to approach every package individually, especially if that's a package I'm not interested at all and only interested in getting the tests to pass.

anton-iskryzhytskyi (approved):

oleksiipiliugin (approved):

vitalyrudenko-affinidi (approved):

PR #199, opened 2021-11-29
[FTL-2719] Refactor migration service

inga-affinidi:


anton-iskryzhytskyi (dismissed):
LGTM

sdk/core/src/services/AffinidiVaultEncryptionService.ts
```diff
@@ -0,0 +1,97 @@
...
+export default class AffinidiVaultEncryptionService {
+  private _keysService
+  private _platformCryptographyTools
+
+  constructor(keysService: KeysService, platformCryptographyTools: IPlatformCryptographyTools) {
+    this._keysService = keysService
+    this._platformCryptographyTools = platformCryptographyTools
+  }
+
+  // TODO: Make this private; it is only used by migration service
```
YuriiSkrypnyk, line 18: This comment looks outdated as I saw you are using that in `AffinidiVaultStorageService`

sdk/core/src/migration/credentials/index.ts
```diff
@@ -18,91 +10,33 @@ interface vcMigrationList {
...
-      }
-
-      const cyphertext = await this.platformCryptographyTools.encryptByPublicKey(publicKeyBuffer, credential)
-
-      encryptedCredentials.push({
-        id: credentialIdHash,
-        types: typeHashes,
-        payload: cyphertext,
-        bloomVaultIndex: credential.bloomId,
-      })
```
YuriiSkrypnyk, line 102: previously we had contract with VMS for such properties and now that is breaking as you are using same method as for AffinidiVaultStorageService.

YuriiSkrypnyk (changes requested):

common-libs/tools-openapi/src/helpers/client.ts
```diff
@@ -62,7 +62,7 @@ type GetRequestOptions<TOperation extends MethodTypeByOperation<any>> = Paramete
 export type GetParams<TOperation extends MethodTypeByOperation<any>> = GetRequestOptions<TOperation>['params']
 
 const parseSpec = (rawSpec: GenericApiSpec) => {
-  const basePath = rawSpec.servers[0].url
+  const basePath = rawSpec.servers[0].url === '/' ? '' : rawSpec.servers[0].url
```
vitalyrudenko-affinidi, line 65: Maybe we should always remove final `/` just in case?
```
rawSpec.servers[0].url.endsWith('/')
    ? rawSpec.servers[0].url.slice(0, -1)
    : rawSpec.servers[0].url
```

vitalyrudenko-affinidi:
Other than existing questions in PR - LGTM

sdk/core/src/migration/credentials/index.ts
inga-affinidi: Because comment for this method (on line 70) says it's "Copy of the private method of `AffinidiVaultStorageService`"?

But apparently it's not just a copy after all; I'll fix it.

common-libs/tools-openapi/src/helpers/client.ts
inga-affinidi: Not sure it is a good idea. What if it will be `/api/v1/`, and the methods themselves will have urls like `users/{id}`? In this case we would want to preserve the slash.
Domain root is a special case.

vitalyrudenko-affinidi: I see, I thought we're consistent with this, but maybe it's not the case.

sdk/core/src/migration/credentials/index.ts
oleksiipiliugin: @inga-affinidi I wrote that comment "Copy of the private method of `AffinidiVaultStorageService`" because I wanted to highlight that it is the same **encryption** implementation as in `AffinidiVaultStorageService` and we can decrypt messages inside `affinidi-vault` using its own decryption methods.

The difference is in the additional field `bloomVaultIndex` that we need for migration process.

inga-affinidi: On the second thought, why do integration tests pass then?

oleksiipiliugin: @inga-affinidi We have integration tests covering only migration endpoints themselves. Another functionality was covered with unit tests.

YuriiSkrypnyk (approved):

anton-iskryzhytskyi (approved):

sdk/core/src/migration/credentials/index.ts
inga-affinidi: @YuriiSkrypnyk , why did you approve this PR then? :)

@YuriiSkrypnyk , @oleksiipiliugin , fixed this issue

oleksiipiliugin (dismissed):

YuriiSkrypnyk (approved):

vitalyrudenko-affinidi (approved):

PR #200, opened 2021-12-03
fix: nuc-173 add wrapper to normalise response from jsonld.frame for vcs with new jsonld context

anton-iskryzhytskyi:
It is temporary solution until we will have time  to find, fix, get PR review in jsonld lib, merge it, publish, then  update jsonld lib in @mattrglobal/json-signatures-bbs,
It does not influence on vcs with old jsonld context, only fix issue with vcs with new type of jsonld context


sdk/node/src/utils/jsonldFrameWrapper.ts
```diff
@@ -0,0 +1,25 @@
+/**
+ * This function is invited to wrap json ld.frame function
+ * in case it returns not a valid segment but @context and @graph fields,
+ * under which in the first index segment with data is contained.
+ * This problem only related to vcs with new jsonld context.
+ * It is temporary solution until we will have time  to find,
+ * fix, get PR review in jsonld lib, merge it, publish,
+ * then  update jsonld lib in @mattrglobal/json-signatures-bbs,
+ * get PR review from them, merge it, publish and only then update this SDK
```
inga-affinidi, line 9: So we cannot create our own jsonld (derived from the original one) because mattrglobal/json-signatures-bbs imports the original jsonld?

sdk/node/src/utils/jsonldFrameWrapper.ts
```diff
@@ -0,0 +1,25 @@
...
+
+const originalJsonldFrame = jsonld.frame
+
+const normalizeFramingResult = (framingResult: Record<string, any>) => {
+  const normalizedFramingResult = framingResult['@graph']
+    ? { ...framingResult, ...framingResult['@graph'][0] }
+    : framingResult
+
+  delete normalizedFramingResult['@graph']
+  return normalizedFramingResult
```
inga-affinidi, line 22: Can't we do
```
const normalizeFramingResult = ({ '@graph': graph, ...restFramingResult }) => ({
  ...restFramingResult,
  ...(graph?.[0]),
})
```
instead?

sdk/node/src/utils/jsonldFrameWrapper.ts
anton-iskryzhytskyi: Nope, we can use force resolution, but it will take much more time for implementation. In the same time, people already want to have this feature. What I see:
* add this monkey patch 
*  then resume work on finding the core of issue => go to jsonld and try to push fix here and in case of failing - create our own solution

oleksiipiliugin (approved):

YuriiSkrypnyk (approved):

cosminm-affnd (approved):

PR #202, opened 2021-12-09
feat: ftl 2682 bbs sd support nested fields

anton-iskryzhytskyi:


anton-iskryzhytskyi:
Waiting for https://github.com/mattrglobal/jsonld-signatures-bbs/pull/161

common-libs/common/src/Affinity.ts
```diff
@@ -600,7 +601,7 @@ export class Affinity {
 
   async deriveSegmentProof<TKeys extends string, TData extends SimpleThing & Record<TKeys, unknown>>(
     credential: VCV1<VCV1Subject<TData>>,
-    fields: TKeys[],
+    fields: string[],
```
inga-affinidi, line 604: Could you please explain this change?
And `TKeys` generic parameter was only needed to put constraints on `credential` and `fields` relationship (`fields` should only contain a subset of non-simple-thing keys in `TData`). If `fields` can contain any strings now, we no longer need `TKeys` parameter at all, we don't even need `TData` probably and can simply write `credential: VCV1<VCV1Subject<SimpleThing>>`. If that's really what we want to do.

inga-affinidi: If we want fields to contain complex paths, we could do it in a type-safe fashion like that (just to give an idea):
```
type InnerPaths<TObject extends Record<string, unknown>> = TObject extends Record<string, unknown> ? {
  // @ts-ignore
  [key in keyof TObject]: key extends string ? `${key}.${Paths<TObject[key]>}` : never
} : never

type Paths<TObject> = TObject extends Record<string, unknown>
    ? ((keyof TObject) | InnerPaths<TObject>[keyof InnerPaths<TObject>])
    : never

async deriveSegmentProof<TData extends SimpleThing>(credential: VCV1<VCV1Subject<TData>>, fields: Paths<TData>) { ... }
```
image

Or just get rid of generic parameters altogether.

common-libs/common/test/unit/utils/objectUtil.test.ts
```diff
@@ -0,0 +1,62 @@
...
+        },
+      }
+
+      const result = injectFieldForAllParentRoots(target, '@explicit', true)
+
+      expect(result).to.deep.equal({
+        address: {
+          buildingNumber: {},
+          country: {},
+          '@explicit': true,
```
inga-affinidi, line 48: Is there any documentation reference for why `address` should contain `@explicit` but `country` should not (even though it is an object)?

common-libs/common/test/unit/utils/objectUtil.test.ts
```diff
@@ -0,0 +1,62 @@
...
+    it('should build object that represents paths to fields', () => {
+      const paths = ['address/buildingNumber', 'address/country', 'id/age', 'id/name/firstName', 'date']
+
+      const representingPathsAsObject = buildObjectSkeletonFromPaths(paths)
+
+      expect(representingPathsAsObject).to.deep.equal({
+        date: {},
+        address: {
+          buildingNumber: {},
+          country: {},
```
inga-affinidi, line 15: Is there any documentation reference for why `country` should be `{}` instead of e.g. `null`?

anton-iskryzhytskyi: To be honest, I haven't try to use not `{}` due to examples contains `{}`
e.g.: [example of revealDoc without nested fields](https://github.com/mattrglobal/jsonld-signatures-bbs/blob/master/sample/ts-node/src/data/deriveProofFrame.json#L11), [example of revealDoc with nested fields](https://github.com/mattrglobal/jsonld-signatures-bbs/blob/a1271750ed5211f1210323534402575967e4fae2/__tests__/__fixtures__/data/test_nested_reveal_document.json#L15)

common-libs/common/test/unit/utils/objectUtil.test.ts
anton-iskryzhytskyi: If a field is a parent field (e.g. contains nested) it should have '@explicit': true flag to indicate only specified fields should be selected. if a field is not a parent -> then it doesn’t make sense to use '@explicit' flag.
I left some links to examples why we use this syntax below, also README.md of jsonld-signatures-bbs https://github.com/mattrglobal/jsonld-signatures-bbs

common-libs/common/src/Affinity.ts
anton-iskryzhytskyi: Template literals, ok, let's try them. Don't see any objections

anton-iskryzhytskyi: It leads to fails during [dependent lib compilation](https://github.com/affinityproject/affinidi-core-sdk/runs/4471050943?check_suite_focus=true)
Also, I can't manage find a way how to cover VCs with nested arrays, e.g.
<details>
  <summary>Example of VC with Arrays (probably HAT)</summary>

```JSON
{
    "type": "HealthPassportBundleCredentialV1",
    "data": {
        "@type": "BundleContainer",
        "fhirVersion": "4.0.1",
        "fhirBundle": {
            "resourceType": "Bundle",
            "@type": "Bundle",
            "entry": [
                {
                    "resource": {
                        "resourceType": "Immunization",
                        "patient": {
                            "reference": "Patient/0"
                        },
                        "vaccineCode": "COVID-19 vaccine, inactivated virus",
                        "manufacturer": {
                            "reference": "Organization/manufacturer-0"
                        },
                        "lotNumber": "SINO-100",
                        "occurrenceDateTime": "2021-06-17T11:30:00+08:00",
                        "doseQuantity": {
                            "value": 1
                        },
                        "performer": {
                            "actor": {
                                "reference": "Organization/administeringCentre-0"
                            }
                        },
                        "protocolApplied": {
                            "seriesDosesPositiveInt": 2
                        }
                    }
                },
                {
                    "resource": {
                        "resourceType": "Immunization",
                        "patient": {
                            "reference": "Patient/0"
                        },
                        "vaccineCode": "COVID-19 vaccine, RNA based",
                        "manufacturer": {
                            "reference": "Organization/manufacturer-1"
                        },
                        "lotNumber": "PB-100",
                        "occurrenceDateTime": "2021-06-17T11:30:00+08:00",
                        "doseQuantity": {
                            "value": 1
                        },
                        "performer": {
                            "actor": {
                                "reference": "Organization/administeringCentre-1"
                            }
                        },
                        "protocolApplied": {
                            "seriesDosesPositiveInt": 2
                        }
                    }
                },
                {
                    "resource": {
                        "resourceType": "Patient",
                        "id": "0",
                        "identifier": [
                            {
                                "value": "S1234567A",
                                "type": {
                                    "text": "National ID"
                                }
                            },
                            {
                                "value": "K1234567Z",
                                "type": {
                                    "text": "Passport ID"
                                }
                            }
                        ],
                        "name": [
                            {
                                "text": "Alice"
                            }
                        ],
                        "gender": "Female",
                        "birthDate": "1990-12-31"
                    }
                },
                {
                    "resource": {
                        "resourceType": "Organization",
                        "id": "manufacturer-0",
                        "name": "Sinovac"
                    }
                },
                {
                    "resource": {
                        "resourceType": "Organization",
                        "id": "manufacturer-1",
                        "name": "Pfizer-BioNTech"
                    }
                },
                {
                    "resource": {
                        "resourceType": "Organization",
                        "id": "administeringCentre-0",
                        "name": "Outram Polyclinic",
                        "address": {
                            "text": "3 Second Hospital Avenue, #02-00 Health Promotion Board Building, Singapore 168937",
                            "country": "Singapore"
                        }
                    }
                },
                {
                    "resource": {
                        "resourceType": "Organization",
                        "id": "administeringCentre-1",
                        "name": "Outram Polyclinic",
                        "address": {
                            "text": "3 Second Hospital Avenue, #02-00 Health Promotion Board Building, Singapore 168937",
                            "country": "Singapore"
                        }
                    }
                }
            ]
        }
    },
    "holderDid": "did:elem:notExist"
}
```
</details>

inga-affinidi: > It leads to fails during dependent lib compilation

But your link points to _test_ errors, not compilation errors?

>  Also, I can't manage find a way how to cover VCs with nested arrays

I just wrote some types according to your code implementation. It seems that all these service methods do not cover nested arrays either?

What are the acceptable values of `fields` in this case?

anton-iskryzhytskyi: > But your link points to _test_ errors, not compilation errors?  

seems wrong pipeline, this one https://github.com/affinityproject/affinidi-core-sdk/runs/4473328218?check_suite_focus=true

> I just wrote some types according to your code implementation. It seems that all these service methods do not cover nested arrays either?
Exactly, it should be covered by jsonld-signatures or, for now, we can write a disclaimer that we don't support picking nested fields in values of arrays. 

> What are the acceptable values of `fields` in this case?
Just anything that a developer decided to pass into this method. And the result could contain null fields. We are putting responsibility for selecting wrong fields on the developer who is using API. If you think it is better to find workaround how to add types that cover everything - let's do it in another PR as it is another story. Main points of this PR are:

anton-iskryzhytskyi: > But your link points to _test_ errors, not compilation errors?  

seems wrong pipeline, this one https://github.com/affinityproject/affinidi-core-sdk/runs/4473328218?check_suite_focus=true

> I just wrote some types according to your code implementation. It seems that all these service methods do not cover nested arrays either?
Exactly, it should be covered by jsonld-signatures or, for now, we can write a disclaimer that we don't support picking nested fields in values of arrays. 

> What are the acceptable values of `fields` in this case?
Just anything that a developer decided to pass into this method. And the result could contain null fields. We are putting responsibility for selecting wrong fields on the developer who is using API. If you think it is better to find workaround how to add types that cover everything - let's do it in another PR as it is another story. Main points of this PR are:
* [cover the issue with BBS signatures ](https://github.com/mattrglobal/node-bbs-signatures/issues/174)
* allow our users to select nested fields of their VCs

inga-affinidi: > We are putting responsibility for selecting wrong fields on the developer who is using API.

But what are the _right_ fields in this case?

cosminm-affnd (approved):

oleksiipiliugin (approved):

common-libs/common/src/Affinity.ts
anton-iskryzhytskyi: it depends on what the VC owner decided to derive. You can find examples in this PR. 
I'll try to improve this, probably add some not strict typing and add tests for VCs with arrays. Let's make it step by step.

inga-affinidi: > You can find examples in this PR.

But these examples do not contain any arrays?

> I'll try to improve this, probably add some not strict typing and add tests for VCs with arrays. Let's make it step by step.

So we are once again merging a PR with changes we do not fully understand ourselves?

If we don't know what are the correct possible values for VCs with arrays, how is VC owner supposed to know that?

anton-iskryzhytskyi: Yes, you are right.
But let's move step by step - these changes are really important as it introduces fixes for security vulnerability.
Also nested selective disclosure is stable for vcs that doesn't have arrays.
For this special case, we have a ticket https://lemmatree.atlassian.net/browse/NUC-180

PR #203, opened 2021-12-13
[CON-100] Implemented didAuthAdapter field

inga-affinidi:


common-libs/internal-api-clients/src/index.ts
```diff
@@ -19,3 +19,5 @@ export {
   CloudWalletApiService,
   VaultMigrationApiService,
 }
+
+export type { DidAuthAdapterType } from '@affinidi/tools-openapi'
```
vitalyrudenko-affinidi, line 23: Is this just for convenience?

dmfilipenko-affinidi (approved):

common-libs/internal-api-clients/src/index.ts
inga-affinidi: To avoid adding unneeded dependency on `tools-openapi` to `wallet-core-sdk` which implements `DidAuthAdapterType`

inga-affinidi:
@vitalyrudenko-affinidi 
> Not really clear to me how is this gonna be used. A description would be helpful!

In Schema Manager frontend, you would be able to create an API client from openapi spec (like it happens in `internal-api-clients`), or maybe add it to `internal-api-clients`; and then create it with didAuthAdapter obtained from wallet.
Instead of having to create did auth adapter manually, and doing all the ceremony with did auth manually, and interacting with schema manager service manually.

PR #204, opened 2021-12-22
[FTL-2763] Don't enforce "data" field in "credentialSubject" when verifying

vitalyrudenko-affinidi:
"data" nesting in VCs is an old way of building the "credentialSubject", that has been used in now deprecated @affinidi/vc-data library

We don't need to enforce its existence anymore, since it's not required by W3C VC Data Model spec and interferes with Schema Manager usage.


common-libs/vc-common/src/vc/v1.ts
```diff
@@ -11,7 +11,7 @@ export type MaybeArray<T> = T | Array<T>
 export type VCV1Subject<Data extends SimpleThing> = {
   '@context'?: string
   id?: string
-  data: Data
+  data?: Data
```
inga-affinidi, line 14: Shouldn't it be
```
export type VCV1Subject<Data extends SimpleThing> = {
  '@context'?: string
  id?: string
} & ({ data: Data } | Data)
```
instead? Updated tests make it seem so.

vitalyrudenko-affinidi: Indeed, this is better. Though, since vc-data is deprecated, I thought these types are no longer needed, but I might be wrong and they might still be useful.

oleksiipiliugin (approved):

anton-iskryzhytskyi (approved):
LGTM

PR #209, opened 2022-03-30
fix: keep Cognito session in case of wrong OTP

rbrazhnyk:


common-libs/user-management/src/CognitoIdentityService.ts
```diff
@@ -196,6 +198,24 @@ export class CognitoIdentityService {
...
+        const { token } = error.inputParams
+
+        try {
+          const parsedToken = JSON.parse(token)
+
+          tempSession[USERNAME] = parsedToken?.logInToken?.Session
+
+        } catch (parseError) {
+          // throw new SdkError('COR-5', {}, parseError)
+          // TODO: How to pass original error?
```
inga-affinidi, line 215: Just throw it. As I understand, `parseError` should not happen?

inga-affinidi: Or just remove the try..catch wrapper

common-libs/user-management/src/CognitoIdentityService.ts
```diff
@@ -196,6 +198,24 @@ export class CognitoIdentityService {
       const cognitoTokens = this._normalizeTokensFromCognitoAuthenticationResult(result.AuthenticationResult)
       return { result: CompleteLoginPasswordlessResult.Success, cognitoTokens }
     } catch (error) {
+      // NOTE: IF below is a fix: when user first time enters wrong OTP, Cognito
+      //       returns new session, for which correct OTP wont work, so user
+      //       will be stuck until server restart and user enters OTP correctly
+      //       from the 1st attempt
+      if (error?.code === 'COR-5' && error?.inputParams?.token) {
```
inga-affinidi, line 205: COR-5 is our error code, as far as I can see there is no way it can be thrown from this `try` block, so this is effectively a no-op?

Also could you please write some integration tests covering this scenario? (They don't have to be in this library, can be in `sdk/core/test/integration/otp` as well)

inga-affinidi: It seems that you need to add your logic to `if (result.ChallengeName === 'CUSTOM_CHALLENGE') {` on line 194 instead of that `catch` block.

common-libs/user-management/src/CognitoIdentityService.ts
```diff
@@ -171,9 +179,9 @@ export class CognitoIdentityService {
 
   async completeLogInPasswordless(token: string, confirmationCode: string): Promise<CompleteLoginPasswordlessResponse> {
     const { Session: tokenSession, ChallengeName, ChallengeParameters } = JSON.parse(token)
-
-    const { USERNAME } = ChallengeParameters
-    const Session = tempSession[USERNAME] || tokenSession
+    //TODO: session is approx 920 character long string
+    const hashedTokenSession = sha256(tokenSession)
```
inga-affinidi, line 183: Why do we need to hash it, instead of using session itself (probably with Map instead of a plain object)?

common-libs/user-management/src/CognitoIdentityService.ts
```diff
@@ -187,7 +195,14 @@ export class CognitoIdentityService {
...
     try {
       const result = await this.cognitoidentityserviceprovider.respondToAuthChallenge(params).promise()
-      tempSession[USERNAME] = result.Session
+      //NOTE : successful OTP return a undefined session . wrong code return a new session
+      if (result.Session) {
+        tempSession[hashedTokenSession] = result.Session
+      } else {
+        //TODO : we still need to think about clean up for sessions that was not finished by user. ex. session was confirmed with wrong pasword 1 or 2 times with out sucess.
+        // potential memory leak.
+        delete tempSession[hashedTokenSession]
```
inga-affinidi, line 204: If we only need to do this when OTP was successful, then probably the better place to do this is right before returning `CompleteLoginPasswordlessResult.Success`

common-libs/user-management/src/CognitoIdentityService.ts
```diff
@@ -198,24 +213,9 @@ export class CognitoIdentityService {
...
-          tempSession[USERNAME] = parsedToken?.logInToken?.Session
-
-        } catch (parseError) {
-          // throw new SdkError('COR-5', {}, parseError)
-          // TODO: How to pass original error?
-          return { result: CompleteLoginPasswordlessResult.ConfirmationCodeWrong }
-        }
-      }
+      // NOTE: not deleted sessions after any errors will block user session
+      delete tempSession[hashedTokenSession]
```
inga-affinidi, line 217: This code will only have an effect in a scenario like this:
1. User initiated passwordless login; cognito returned session A;
2. User entered a wrong OTP; we sent request to cognito with wrong OTP and session A; cognito returned session B;
3. User entered some OTP; we sent request to cognito with that new OTP and session B; there was an error from cognito (e.g. timeout or something).

Does this mean that when user retries the request (or enters some third OTP), we should use session A instead of B? Why?

inga-affinidi (changes requested):
Please revert these huge changes to package-locks, this MR shouldn't have 207 thousands changed lines

common-libs/user-management/src/CognitoIdentityService.ts
Volland: session is almost 1kb long . so it is to expensive to use it as a key

common-libs/user-management/src/CognitoIdentityService.ts
Volland: this code will have effect on any error i already add integration test cases 
so cases 
1) user entered OTP wrongly 3 times 
2) user entered OTP once and that OTP expired 
3) any failure will block a session in a map forever

common-libs/user-management/src/CognitoIdentityService.ts
inga-affinidi: But we're removing it after successful login, and we already cache DID documents etc, I don't think having a session in memory (only until it's removed with a successful login) will cause us any troubles.
Or if it will, let's fix them later. Right now it feels like increased complexity (and potential incompatibility with the browser environment as you yourself mentioned) for the sake of premature optimization.

common-libs/user-management/src/CognitoIdentityService.ts
inga-affinidi: So if cognito was not available, an user should attempt to log in again instead of retrying confirmation with their existing session.

sdk/core/test/integration/otp/core.test.ts
```diff
@@ -352,6 +358,112 @@ parallel('CommonNetworkMember [OTP]', () => {
...
+      try {
+        await AffinidiWallet.completeLogInPasswordless(options, loginToken, randomOTP())
+      } catch (errFirstTry) {
+        expect(errFirstTry).to.be.instanceOf(SdkError)
+        expect(errFirstTry.name).to.eql('COR-5')
+      }
+
+      try {
+        await AffinidiWallet.completeLogInPasswordless(options, loginToken, randomOTP())
+      } catch (errFirstTry) {
```
inga-affinidi, line 375: so if there is no exception, test will succeed

sdk/core/test/integration/otp/core.test.ts
```diff
@@ -352,6 +358,112 @@ parallel('CommonNetworkMember [OTP]', () => {
...
+      }
+
+      await wait(180)
+
+      const commonNetworkMember = await AffinidiWallet.completeLogInPasswordless(options, loginToken, loginCode)
+      checkIsWallet(commonNetworkMember)
+      expect(commonNetworkMember.did).to.exist
+    })
+
+    it('user get COR-13 on 3 wrong otp codes ', async function () {
```
inga-affinidi, line 387: But we already have `it('throws COR-13 at the third call to #confirmSignIn with wrong confirmation code'` test here

sdk/core/test/integration/otp/core.test.ts
```diff
@@ -352,6 +358,112 @@ parallel('CommonNetworkMember [OTP]', () => {
...
+      const newLoginToken = await AffinidiWallet.initiateLogInPasswordless(options, inbox.email, messageParameters)
+      const newloginCode = await waitForOtpCode(inbox)
+
+      const newNetworkMember = await AffinidiWallet.completeLogInPasswordless(options, newLoginToken, newloginCode)
+      checkIsWallet(newNetworkMember)
+      expect(newNetworkMember.did).to.exist
+    })
+
+    // NUC-269
+    it('user is able to login with a new session even if old was with wrong otp', async function () {
```
inga-affinidi, line 446: Also probably makes sense to test the scenario where an user logged in successfully, and then attempts to log in again

sdk/core/test/integration/otp/core.test.ts
Volland: yep but here we make sure that user still able to login . so before test was there but we have seriose bug there . so i reinsure that user are able to do a passwordless login after getting errors and etc

common-libs/user-management/src/CognitoIdentityService.ts
Volland: Not really we have a cases when user enter a wrong OTP and never complete a flow . so it will grow 2x in memory over time.

common-libs/user-management/src/CognitoIdentityService.ts
Volland: we just do a clean up and avoiding a dead record there

common-libs/common/src/services/KeysService.ts
```diff
@@ -9,6 +8,7 @@ import { convertDecryptedSeedBufferToString, parseDecryptedSeed } from '../share
 import DigestService from './DigestService'
 import DidDocumentService from './DidDocumentService'
 
+const createHash = require('create-hash/browser')
```
anton-iskryzhytskyi, line 11: why we use require? Is it not outdated? There are not type support with require,  as I know , that could lead to mistakes and typos.
Also, what happen if this code will be used on backend side?

common-libs/user-management/src/CognitoIdentityService.ts
```diff
@@ -1,3 +1,4 @@
+const createHash = require('create-hash/browser')
```
anton-iskryzhytskyi, line 1: why we use "require" here instead of "import"?

sdk/core/test/integration/otp/core.test.ts
```diff
@@ -352,6 +358,143 @@ parallel('CommonNetworkMember [OTP]', () => {
...
+      try {
+        await AffinidiWallet.completeLogInPasswordless(options, loginToken, randomOTP())
+      } catch (errFirstTry) {
+        expect(errFirstTry).to.be.instanceOf(SdkError)
+        expect(errFirstTry.name).to.eql('COR-5')
+        expect(errFirstTry.context.newToken).to.exist
+      }
+
+      try {
+        await AffinidiWallet.completeLogInPasswordless(options, loginToken, randomOTP())
```
anton-iskryzhytskyi, line 402: the same as above

sdk/core/test/integration/otp/core.test.ts
```diff
@@ -352,6 +358,143 @@ parallel('CommonNetworkMember [OTP]', () => {
...
+      } catch (errFirstTry) {
```
anton-iskryzhytskyi, line 411: the same as above

sdk/core/test/integration/otp/core.test.ts
```diff
@@ -352,6 +358,143 @@ parallel('CommonNetworkMember [OTP]', () => {
...
+      expect(commonNetworkMember.did).to.exist
+    })
+
+    it('use newToken session from error for existing OTP', async function () {
+      const { inbox } = await createUser()
+      const loginToken = await AffinidiWallet.initiateLogInPasswordless(options, inbox.email, messageParameters)
+      const loginCode = await waitForOtpCode(inbox)
+      try {
+        await AffinidiWallet.completeLogInPasswordless(options, loginToken, randomOTP())
+      } catch (errFirstTry) {
```
anton-iskryzhytskyi, line 431: the same as above

sdk/core/test/integration/otp/core.test.ts
```diff
@@ -352,6 +358,143 @@ parallel('CommonNetworkMember [OTP]', () => {
...
+        expect(errFirstTry).to.be.instanceOf(SdkError)
+        expect(errFirstTry.name).to.eql('COR-5')
+        expect(errFirstTry.context.newToken).to.exist
+      }
+
+      let newToken
+      try {
+        await AffinidiWallet.completeLogInPasswordless(options, loginToken, randomOTP())
+      } catch (errSecondTry) {
```
anton-iskryzhytskyi, line 440: the same as above

sdk/core/test/integration/otp/core.test.ts
```diff
@@ -352,6 +358,143 @@ parallel('CommonNetworkMember [OTP]', () => {
...
+      const { inbox } = await createUser()
+      const loginToken = await AffinidiWallet.initiateLogInPasswordless(options, inbox.email, messageParameters)
+      const loginCode = await waitForOtpCode(inbox)
+
+      const commonNetworkMember = await AffinidiWallet.completeLogInPasswordless(options, loginToken, loginCode)
+      checkIsWallet(commonNetworkMember)
+      expect(commonNetworkMember.did).to.exist
+      try {
+        await AffinidiWallet.completeLogInPasswordless(options, loginToken, loginCode)
+      } catch (errFirstTry) {
```
anton-iskryzhytskyi, line 462: the same as above

anton-iskryzhytskyi (commented):
I left some comments related to tests and several questions about "require",  from other sides looks good for me, good job!

common-libs/common/src/services/KeysService.ts
inga-affinidi: Also I don't see any reason for `KeysService` to be affected by this PR.

common-libs/url-resolver/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# release 1.1.3 (2022-04-04)
+Remove unused service 
+  - user auth 
+  - sff-issuer
```
inga-affinidi, line 4: Can we keep it to a separate MR?

common-libs/user-management/src/UserManagementService.ts
```diff
@@ -201,7 +201,7 @@ export class UserManagementService {
       case CompleteLoginPasswordlessResult.ConfirmationCodeExpired:
         throw new SdkErrorFromCode('COR-17', { confirmationCode })
       case CompleteLoginPasswordlessResult.ConfirmationCodeWrong:
-        throw new SdkErrorFromCode('COR-5')
+        throw new SdkErrorFromCode('COR-5', { newToken: response.token })
```
inga-affinidi, line 204: So are we returning new session or caching it?

inga-affinidi (dismissed):
This MR still has 207k changed lines...

common-libs/common/src/services/KeysService.ts
Volland: we have a discussion with @DenisPopov15 . it is a bug in sdk so all functions that use createhash will work only in nidejs.

Volland: so what we should do ? just ignore issue ?

sdk/core/test/integration/otp/core.test.ts
```diff
@@ -352,6 +358,143 @@ parallel('CommonNetworkMember [OTP]', () => {
...
+      checkIsWallet(commonNetworkMember)
+      expect(commonNetworkMember.did).to.exist
+    })
+
+    it('user get COR-13 on 3 wrong otp codes ', async function () {
+      const { inbox } = await createUser()
+      const loginToken = await AffinidiWallet.initiateLogInPasswordless(options, inbox.email, messageParameters)
+      await waitForOtpCode(inbox)
+      try {
+        await AffinidiWallet.completeLogInPasswordless(options, loginToken, randomOTP())
```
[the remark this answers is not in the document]
Volland: add expectation

sdk/core/test/integration/otp/core.test.ts
Volland: add expectation

sdk/core/test/integration/otp/core.test.ts
Volland: add expectation

common-libs/user-management/src/CognitoIdentityService.ts
Volland: current library do not support types for a browser module . i ll make a separate jira to find a better solution

common-libs/url-resolver/CHANGELOG.md
Volland: nop tests are not running any more

common-libs/user-management/src/UserManagementService.ts
Volland: local in memory object do not work with cloud-wallet or any server side usage . so we return a session and client is responsible to send a correct session back . I add a proper section to a read.me of core library

sdk/core/test/integration/otp/core.test.ts
Volland: added failure expectation

sdk/core/test/integration/otp/core.test.ts
Volland: added failure expectation

sdk/core/test/integration/otp/core.test.ts
Volland: I dont get the point of additional tests . what we are trying to achieve with this test ?

anton-iskryzhytskyi (approved):
LGTM

DenisPopov15 (approved):

common-libs/common/src/services/KeysService.ts
inga-affinidi: And after this fix they are only going to work in browser?
We should have removed this change from this MR and create a separate ticket and fix it _properly_.

common-libs/url-resolver/CHANGELOG.md
inga-affinidi: But they did just a week ago, did something change?

common-libs/user-management/src/CognitoIdentityService.ts
inga-affinidi: So in the end we're cleaning it up now in most cases.
And in the remaining cases (e.g. user entered a wrong OTP and never completed the flow) we still have a memory leak with the current approach, except that we leak 32 bytes (+C) instead of a kilobyte (+C) per abandoned session.

This is not going to solve memory leaks, only make them smaller somewhat.
And I don't see the point in increasing complexity in making these memory leaks smaller until we are certain that their size is causing issues.
Maybe even with a kilobyte per session we wouldn't have any problems.
Or maybe even with 32 bytes per session we would have serious problems.

Also we're caching DIDs currently, which are much larger, and this was not causing us any trouble so far.

common-libs/user-management/src/CognitoIdentityService.ts
inga-affinidi: We probably need to communicate this fact to clients somehow?

> So if cognito was not available, an user should attempt to log in again instead of retrying confirmation with their existing session.

Currently it might be expected that if an attempt to confirm login returned some unknown error we can retry to confirm login again.
As I understand, with the current implementation we won't be able to confirm login anymore (we'll keep getting unknown errors probably?), so we need to restart the login flow.
Is my understanding correct?

common-libs/user-management/src/UserManagementService.ts
inga-affinidi: So we have to update contracts of cloud-wallet-api as well; and in-memory storage only remains for backwards compatibility with SDK consumers that are not aware of these changes, correct?

sdk/core/test/integration/otp/core.test.ts
inga-affinidi: To cover a scenario which (as we assume) did not work prior to these changes, to make sure it works now.

PR #210, opened 2022-04-08
feat: NUC-271 fix password less flow for users with arbitrary name and changed email

Volland:
Fix for NYC-271.
Bug reported by HR wallet team 
- create user with arbitrary name 
- change user email 
- use the same email for signIn 
So as result sign in flow do not identify user and do sign Up that will fail with aliase error failure 
Expected behavior - we identify that confirmed user already exist and do a login flow.

After quick investigation we identified that normalising of user name for confirmed user is a root cause.

### DIFF
--- a/common-libs/internal-api-clients/package-lock.json
+++ b/common-libs/internal-api-clients/package-lock.json
@@ -1,12 +1,12 @@
 {
 	"name": "@affinidi/internal-api-clients",
-	"version": "1.2.0",
+	"version": "1.2.1",
 	"lockfileVersion": 2,
 	"requires": true,
 	"packages": {
 		"": {
 			"name": "@affinidi/internal-api-clients",
-			"version": "1.2.0",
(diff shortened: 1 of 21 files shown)


anton-iskryzhytskyi:
@Volland looks good except package-lock also should have changed version as package.json, as I know.

dmfilipenko-affinidi (approved):

Volland:
@anton-iskryzhytskyi last time  i was asken not to commit lock files  i ll add them

dmfilipenko-affinidi:
Also need to update changelog for user-managment lib

anton-iskryzhytskyi (approved):

anton-iskryzhytskyi (approved):

dmfilipenko-affinidi (approved):

dmfilipenko-affinidi (approved):

rbrazhnyk (approved):

PR #219, opened 2022-06-22
fix(FTL-5605): fixed BigInt error for react native

oleksiipiliugin:
Fixed BigInt error for react native. Freezing `jsontokens` version at 3.0.0

Big Int [does not work](https://github.com/facebook/react-native/issues/28492) at react native.
In SDK we use `jsontokens` dependencies to work with `JWT`.
In this [PR](https://github.com/stacks-network/jsontokens-js/pull/63) jsontoken's maintainers added new dependencies(@noble/hashes @noble/secp256k1) for crypto and hashing methods. @noble/hashes, @noble/secp256k1 are fastest decisions among other JS libs, but it contains [BigInt](https://github.com/paulmillr/noble-secp256k1/blob/main/index.ts#L16) which is not suitable for react native.

- [x] changelog and versions


common-libs/common/CHANGELOG.md
```diff
@@ -1,3 +1,5 @@
+# release 2.1.6 (2022-06-22)
+* updated `internal-api-clients`, `tools-common`
```
Volland, line 2: @oleksiipiliugin it is not clean why we update a version of lib could we mention rootcause everywhere in a change log ?

Volland (approved):

Volland (approved):

cosminm-affnd (approved):

PR #221, opened 2022-06-23
Draft: [Feature] ftl-5589 skip bloom vault

Volland:


sdk/core/src/dto/shared.dto.ts
```diff
@@ -106,6 +106,18 @@ export class SdkOptions {
...
   @IsOptional()
   issueSignupCredential?: boolean
+
+  @IsBoolean()
+  @IsOptional()
+  queryBloomVault?: boolean
+
+  @IsString()
+  @IsOptional()
+  userPoolId?: string
```
cosminm-affnd, line 116: minor: Shouldn't these (userPoolId & clientId) be under a cognito object? We may have other clientId in the future

cosminm-affnd (approved):

sdk/core/src/services/WalletStorageService.ts
```diff
@@ -171,12 +180,14 @@ export default class WalletStorageService {
     try {
       return await this._affinidiVaultStorageService.getCredentialById(credentialId, storageRegion)
     } catch (error) {
+      // should be deleted during migration to affinidi-vault Phase #2
```
anton-iskryzhytskyi, line 183: looks like a bug
```
    try {
      return await this._affinidiVaultStorageService.getCredentialById(credentialId, storageRegion)
    } catch (error) {
      // should be deleted during migration to affinidi-vault Phase #2
      # condition is true when error AVT-2 and queryBloomVault = true
     # condition is false when error AVT-2 and queryBloomVault = false
      if (error.code === 'AVT-2' && this._queryBloomVault) {
        return await this._bloomVaultStorageService.getCredentialById(credentialId, storageRegion)
      }
      # throws all errors exclude AVT-2, but there are a chance when queryBloomVault is false and error is AVT-2
     # so when queryBloomVault is false  all AVT-2 ignored
      if (error.code !== 'AVT-2') {
        throw error
      }
    }
```

sdk/core/src/services/WalletStorageService.ts
```diff
@@ -204,10 +217,12 @@ export default class WalletStorageService {
...
-    try {
-      await this._bloomVaultStorageService.deleteAllCredentials(storageRegion)
-    } catch (error) {
-      throw new SdkErrorFromCode('COR-0', {}, error)
+    if (this._queryBloomVault) {
+      try {
+        await this._bloomVaultStorageService.deleteAllCredentials(storageRegion)
+      } catch (error) {
+        throw new SdkErrorFromCode('COR-0', {}, error)
+      }
```
anton-iskryzhytskyi, line 225: I'd recommend moving this part into separate function to reduce the amount of nested code, before these changes it was 1 nested block - now 2

sdk/core/test/unit/services/WalletStorageService.test.ts
```diff
@@ -34,7 +34,7 @@ let encryptedSeed: string
 
 const region = 'eu-west-2'
 
-const createWalletStorageService = () => {
+const createWalletStorageService = (queryBloomVault?: boolean) => {
```
anton-iskryzhytskyi, line 37: BTW why it doesn't have default option? 
`const createWalletStorageService = (queryBloomVault: boolean = true) => {`
it could then simplify this block of code
`queryBloomVault: queryBloomVault === false ? false : true,`
to 
`queryBloomVault`

PR #228, opened 2022-08-01
fix(versions): updated versions and changelog

oleksiipiliugin:
Follow-up changes for `vc-data`, `vc-common`. Updated dependencies versions. Added changes descriptions to changelogs.
FTL-6271 - Fixed did document validation error.


Volland (approved):
LGTM

PR #229, opened 2022-08-09
feat(FTL-5593): add origin param when requesting Registry

dhuan-affinidi:


common-libs/internal-api-clients/src/services/RegistryApiService.ts
```diff
@@ -24,7 +24,16 @@ export default class RegistryApiService {
   }
 
   async anchorDid(params: GetParams<typeof clientMethods.AnchorDid>) {
-    return this.client.AnchorDid({ params })
+    const originFromEnv = process.env.REGISTRY_ORIGIN
```
Volland, line 27: @dhuan-affinidi  do we have a sdk option for origin ? why we have a need for env var on core dsk code? it could be a challenge for react native users.

PR #235, opened 2022-08-17
fix(FTL-6435): fixed `package.json` module after refactored build pro…

oleksiipiliugin:
Fixed `package.json` module after refactored build process


anton-iskryzhytskyi (approved):

sdk/vc-data/tsconfig.json
```diff
@@ -21,6 +21,7 @@
     "moduleResolution": "node",
     "jsx": "react",
     "esModuleInterop": true,
-    "outDir": "dist"
+    "outDir": "dist",
+    "skipLibCheck": true
```
Volland, line 25: @oleksiipiliugin why we add skipLibCheck?

Volland (approved):

sdk/vc-data/tsconfig.json
oleksiipiliugin: pls see [docs](https://www.typescriptlang.org/tsconfig#skipLibCheck)

PR #250, opened 2022-10-18
feat: [FTL-7231] random username

anton-iskryzhytskyi:


yaroslava-kurash (approved):

common-libs/user-management/src/UserManagementService.ts
```diff
@@ -177,17 +194,25 @@ export class UserManagementService {
 
   async completeSignUpForEmailOrPhone(token: string, confirmationCode: string) {
     const { login, shortPassword } = this.parseSignUpToken(token)
-    this._loginShouldBeEmailOrPhoneNumber(login)
```
Volland, line 180: @anton-iskryzhytskyi why we remove this check?

common-libs/user-management/src/UserManagementService.ts
```diff
@@ -177,17 +194,25 @@ export class UserManagementService {
...
 
-  async doesUnconfirmedUserExist(username: string) {
-    const normalizedUsername = normalizeUsername(username)
-    return this._cognitoIdentityService.doesUnconfirmedUserExist(normalizedUsername)
+  /**
+   * Uses an ep on KeysService to check if record exists or not for given username or alias (email / phone)
+   * @param login
+   */
+  async doesUnconfirmedUserExist(login: string) {
+    const { userExists, isUnconfirmed } = await this.doesUserExist(login)
```
Volland, line 207: @anton-iskryzhytskyi, how we know that user is unconfirmed? As i understand, we query users by attributes but do we have a way to check the confirmation status?
We could check attributes or do a doesConfirmedUser exist call in pair with this method

anton-iskryzhytskyi: to answer on your question I'd like to send you to the implementation of AWB there
`const isUnconfirmed = userExists && !(user.UserStatus === 'CONFIRMED')`

common-libs/user-management/src/UserManagementService.ts
anton-iskryzhytskyi: Haw it was:
signUp token had a format 'login::password', then we derive username from login, we need username to complete sign-up
Now:
signUp token has a format 'username::password', username is uuid, so can't be email or phone

common-libs/user-management/src/UserManagementService.ts
Volland: oki got it

Volland (approved):
looks good

PR #267, opened 2022-11-14
feat(FTL-7654): added update to SDK to support the `Truecaller` flow

oleksiipiliugin:
Implementation of the signin flow with the [Truecaller](https://docs.truecaller.com/truecaller-sdk/) profile/token.

Logic is the same as for sigin passwordless with phone number, but without sending OTP to the user.
With this new feature, we are able to create a new user in the Cognito pool or login if it exists.
The main changes are in `sdk/core` and `lib/user-management`. Also, support for new endpoints was added for the internal key storage service.


yaroslava-kurash (commented):
Shouldn't versions of `@affinidi/wallet-browser-sdk`  `@affinidi/wallet-expo-sdk`  `@affinidi/wallet-node-sdk` and `@affinidi/wallet-react-native-sdk` be updated too as they use new version of `@affinidi/wallet-core-sdk`

yaroslava-kurash (approved):

sdk/core/test/integration/index.test.ts
```diff
@@ -4,8 +4,8 @@ import './env'
 import { useNodeFetch } from '@affinidi/platform-fetch-node'
 useNodeFetch()
 
-require('./services')
-require('./AffinityWallet.test')
-require('./NetworkMember.test')
+// require('./services')
```
linn-affinidi, line 7: forgot to enable back?

sdk/core/test/integration/otp/core.test.ts
```diff
@@ -643,3 +658,79 @@ parallel('CommonNetworkMember [OTP]', () => {
     })
   })
 })
+
+// describe('CommonNetworkMember [TrueCaller]', () => {
```
Volland, line 662: @oleksiipiliugin why we comment it out ?

Volland: ah we probably need to deploy awb first

Volland (approved):

yaroslava-kurash (approved):

PR #269, opened 2022-11-16
feat: [FTL-7647] User creation timestamp

linn-affinidi:
- [x] Add adminGetUserInfo() to common-libs/internal-api-clients
- [x] Expose user creation timestamp to client like `did`


sdk/core/src/CommonNetworkMember/NetworkMemberWithCognito.ts
```diff
@@ -24,17 +24,20 @@ import { normalizeShortPassword } from '../shared/normalizeShortPassword'
 
 type UserDataWithCognito = ConstructorUserData & {
   cognitoUserTokens: CognitoUserTokens
+  createdDate: string
```
Volland, line 27: we dont need to add it as a field or keep this data . we could just use  a call to service .

Volland: Idea was to have it as a method on sdk

common-libs/internal-api-clients/src/services/KeyStorageApiService.ts
```diff
@@ -82,6 +82,11 @@ export default class KeyStorageApiService {
     return this.client.AdminLogOutUser({ authorization: accessToken })
   }
 
+  async adminGetUserInfo({ accessToken }: { accessToken: string }) {
+    const result = await this.client.AdminGetUserInfo({ authorization: accessToken })
```
oleksiipiliugin, line 86: Just FYI
it is better to use auto-generated types here next time:
```
async adminGetUserInfo(params: GetParams<typeof clientMethods.AdminGetUserInfo>) {
    const result = await this.client.AdminGetUserInfo(params)
```

Volland (approved):

common-libs/internal-api-clients/src/services/KeyStorageApiService.ts
linn-affinidi: sure, will change it next time.

PR #273, opened 2022-12-06
feat: [CFR-275] allow resolve elem legacy on client

anton-iskryzhytskyi:
Allows the SDK users to resolve legacy elem did without requesting the registry


Volland (commented):
looks good but tests is missed

PR #275, opened 2022-12-08
fix: FTL-7826 Return valid credentials

linn-affinidi:
Return valid creds are returned if there is one corrupted for some users.


sdk/core/test/unit/services/WalletStorageService.test.ts
```diff
@@ -149,6 +171,22 @@ describe('WalletStorageService', () => {
     })
   })
 
+  describe('#getAllCredentialsWithBadEncryption', () => {
```
anton-iskryzhytskyi, line 174: I'd like to keep `describe` block with method name only, and move this detail (BadEncryption) to the `it` block

So please move you it block under `#getAllCredentials`

sdk/core/test/unit/services/WalletStorageService.test.ts
```diff
@@ -149,6 +171,22 @@ describe('WalletStorageService', () => {
...
+    it('should filter bad credential from Affinidi Vault without throwing error', async () => {
+      sinon
+        .stub(AffinidiVaultApiService.prototype, 'searchCredentials')
+        .withArgs(region)
+        .callsFake(function anon(): any {
+          return { body: { credentials: [encryptedBadCredential] } }
```
anton-iskryzhytskyi, line 180: WDYT to extend this list of credentials with good ones?
E.g. Good {id=1} Bad {id=2} Good {id=3} Bad {id=4} Good {id=5}
and then compare the result contains only `id in 1, 3, 5`

anton-iskryzhytskyi:
LGTM! small suggestion about tests only

sdk/browser/CHANGELOG.md
```diff
@@ -1,3 +1,5 @@
+# release 7.9.0 (2022-12-09)
```
anton-iskryzhytskyi, line 1: WDYT to use patch update there, as this changes doesn't introduce any breaking changes (major), no new feature (minor)?

sdk/browser/CHANGELOG.md
```diff
@@ -1,3 +1,5 @@
...
+fix: filter corrupted credentials from getCredentials()
```
anton-iskryzhytskyi, line 2: Please follow the previous approach (e.g. added/updated/fixed), I agree it is discussable to have commit message, existing approach or something else, but if everyone use his own way to tell about changes - it leads to mess

anton-iskryzhytskyi (changes requested):
I'd like to keep one approach about changelog, and versions (major - BREAKING/minor-feature/patch-fix).

sdk/browser/CHANGELOG.md
linn-affinidi: ok, i will use patch.

sdk/browser/CHANGELOG.md
linn-affinidi: ok

sdk/core/test/unit/services/WalletStorageService.test.ts
linn-affinidi: ok

sdk/core/test/unit/services/WalletStorageService.test.ts
linn-affinidi: the limitation is - i can only create walletStorageService with either `createWalletStorageService` or `createWalletStorageServiceWithBadEncryption`. The main goal of the test is to return array without throwing exception. 

Yes, test can be extended but I prefer to find the actual root cause and fix it. By right, no credential should be corrupted and fail to decrypt (since same sdk is used to encrypt)

anton-iskryzhytskyi (approved):
LGTM

oleksiipiliugin (approved):

PR #281, opened 2023-01-20
fix: [FTL-8722] upgrade jsonwebtoken and class-validator

yaroslava-kurash:


yaroslava-kurash:
`sdk/core/package-lock.json` has 23k changes because indentation was changed after `npm i`

PR #284, opened 2023-01-26
fix(FTL-8754): implemented Cognito Lambda error handling

oleksiipiliugin:
- `CompleteLoginPasswordlessResult` was [extended](https://github.com/affinidi/affinidi-core-sdk/pull/284/files#diff-fb96193d1cb2d478423b97182b3578d3965f49d08a8104e5d4465af339fbad7fR46) with `VerifyAuthLambdaCustomError` in order to handle Cognito Lambda error handling


sdk/core/test/integration/otp/core.test.ts
```diff
@@ -733,5 +733,20 @@ parallel('CommonNetworkMember [OTP]', () => {
 //       const { wallet } = await AffinidiWallet.signInWithProfile(options, trueCallerTestProfile)
 //       checkIsWallet(wallet)
 //     })
+//
```
anton-iskryzhytskyi, line 736: why it is commented?

anton-iskryzhytskyi (approved):
LGTM, minor question

sdk/core/test/integration/otp/core.test.ts
oleksiipiliugin: it is temporarily blocked by infrastructure issues. will be uncommented after it will be resolved

linn-affinidi: if it is temp, i think we should use it.skip()

PR #288, opened 2023-01-30
fix: return getTrueCallerUrl method to TrueCallerService

yaroslava-kurash:


common-libs/user-management/src/TrueCallerPublicKeyManager.ts
```diff
@@ -8,7 +8,7 @@ type TrueCallerPK = { keyType: string; key: string }
 export class TrueCallerPublicKeyManager {
```
anton-iskryzhytskyi, line 8: please also export `TrueCallerPublicKeyManager` from user-management lib

Volland (approved):

oleksiipiliugin (approved):

PR #303, opened 2023-03-07
feat(FTL-9214): adopt core SDK to use external keys providers

anton-iskryzhytskyi:
These changes allow setting any implementation of KeyManager for SDK, e.g. Affinidi KMS, or some crypto wallet


linn-affinidi (approved):
LGTM

common-libs/common/src/services/KeyManager/KeyManager.ts
```diff
@@ -0,0 +1,46 @@
...
+   */
+  sign?(buffer: Buffer): Buffer
+
+  /**
+   * Signs a digest
+   * @param buffer
+   */
+  signAsync(buffer: Buffer): Promise<Buffer>
+
+  getAnchorTransactionPublicKey(): Promise<string>
```
linn-affinidi, line 38: this could be sync method and implement as property `get AnchorTransactionPublicKey()`?

anton-iskryzhytskyi: I'd say no (not sure for 100%), but if this key will be given by external provider through network (e.g. external provider Aff KMS) - I think it is safe to leave async.

linn-affinidi (approved):

PR #320, opened 2023-08-23
fix: did & added validation on expiresAt

mritunjayk2:


DenisPopov15 (approved):

common-libs/common/src/Affinity.ts
```diff
@@ -86,8 +86,14 @@ export class Affinity {
       throw new Error('Signature on token is invalid')
     }
 
-    if (payload.exp < Date.now()) {
-      throw new Error('Token expired')
+    if (payload.expiresAt) {
+      if (payload.expiresAt < Date.now()) {
+        throw new Error('Token expired')
+      }
+    } else {
```
cosminm-affnd, line 93:
```suggestion
    } else if (payload.exp) {
```

Is it possible to have no expiration field?

DenisPopov15 (changes requested):
we should have at least some exp check always

